PR #1, opened 2015-07-24
Adding VW JNI external models.

deaktator:
The CLI for creating VW models uses the `--external` flag to determine whether an embedded or externalized model should be produced.  Remember that when using `--external`, the location refers to the location at model instantiation time, not model JSON creation time.

Look at the unit tests.  The 'modelUrl' field in the JSON for VwJniModel can have many different protocols.  The following VFS file systems will not cause the content to be copied to a local temp file:
- [File](https://commons.apache.org/proper/commons-vfs/filesystems.html#Local_Files)
- [RES](https://commons.apache.org/proper/commons-vfs/filesystems.html#res)
- No Protocol (_same as [File](https://commons.apache.org/proper/commons-vfs/filesystems.html#Local_Files) VFS file system_)

The following are known to copy to local: 
- [Temp](https://commons.apache.org/proper/commons-vfs/filesystems.html#Temporary_File)
- [RAM](https://commons.apache.org/proper/commons-vfs/filesystems.html#ram)
- [_GZIP_ and _BZIP2_](https://commons.apache.org/proper/commons-vfs/filesystems.html#gzip_and_bzip2)

The following file systems should copy to local but have not been tested: 
- FTP
- FTPS
- HDFS
- HTTP
- HTTPS
- Jar
- SFTP
- Tar
- Zip

### DIFF
--- a/aloha-cli/src/test/scala/com/eharmony/aloha/cli/CliTest.scala
+++ b/aloha-cli/src/test/scala/com/eharmony/aloha/cli/CliTest.scala
@@ -45,6 +45,8 @@ class CliTest extends TestWithIoCapture(CliTest) {
               |        numeric id of the model.
               |  --vw-args <value>
               |        arguments to vw
+              |  --external
+              |        link to a binary VW model rather than embedding it inline in the aloha model.
               |  --num-missing-thresh <value>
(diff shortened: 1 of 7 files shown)


deaktator:
Updated the code to automatically delete the temp files as quickly as possible.

PR #33, opened 2015-08-19
This should address issue 32

jmorra:
This should address the issue with using Vfs2 as the only way to read embedded models.  @deaktator please review this.  Note that I didn't add any tests for this I just made all the current tests pass.  I also forced the writing of the via flag in all models to make the model more explicit but I did not remove Vfs2 as the default when reading if via is absent.


deaktator:
Awesome. This works for you on spark, and on hadoop, no?

jmorra:
Yup.

PR #36, opened 2015-08-28
6 model descriptions

deaktator:
Would like to merge docs


src/site/markdown/model_formats.md
```diff
@@ -0,0 +1,1996 @@
...
+  "features": { "intercept": "intercept" },
+  "weights": { "intercept": 0.5 }
+}
+```
+
+There is a little bit going on here behind the scenes.  The `intercept` function that appears as the sole *value* in 
+the `features` map is a special function that is included when importing `com.eharmony.aloha.feature.BasicFuncions._`. 
+It is in the `com.eharmony.aloha.feature.Intercept` trait that is mixed into `BasicFuncions`.  What this function does
+is create one key-value pair: `("", 1.0)`.  Then, as has been mentioned before, the feature name is prepended to the 
+key, so the final key-value pair is `("intercept", 1.0)`.  Then the associated weight (which is the
```
jmorra: I think you need to finish this thought.

src/site/markdown/model_formats.md
```diff
@@ -0,0 +1,1996 @@
...
+
+```xml
+<dependency>
+  <groupId>com.eharmony</groupId>
+  <artifactId>aloha-vw-jni</artifactId>
+  <version>...</version>
+<dependency> 
+```
+ 
+VW does not need to be installed on the computer where the Aloha VW model is used.  The VW JNI library is contains
```
jmorra: The VW JNI library contains

src/site/markdown/model_formats.md
```diff
@@ -0,0 +1,1996 @@
...
+`modelType` field must be `VwJNI`.
+
+#### (VW) features
+
+See [Regression model features](#aR_features) for details.
+
+#### (VW) vw
+
+The `vw` Object has two main components.  The first component is a representation of the binary VW model, or 
+"*initial regressor*" in VW terminology.  This is the model parameterization.  The second component is set of
```
jmorra: The second component is the set of

src/site/markdown/model_formats.md
```diff
@@ -0,0 +1,1996 @@
...
+
+#### (VW) features
+
+See [Regression model features](#aR_features) for details.
+
+#### (VW) vw
+
+The `vw` Object has two main components.  The first component is a representation of the binary VW model, or 
+"*initial regressor*" in VW terminology.  This is the model parameterization.  The second component is set of 
+parameters used by VW.  This includes things like link function, prediction ranges, skip grams, namespace
```
jmorra: This includes things like the link function

src/site/markdown/model_formats.md
```diff
@@ -0,0 +1,1996 @@
...
+</table>
+
+##### (VW) vw creationTime
+
+`creationTime` is optional and is used when copying the content of the binary VW model to the local disk on the 
+computer where the Aloha model is run.  When no value is supplied, the value will be set to
+`java.lang.System.currentTimeMillis()`.
+
+##### (VW) vw model
+
```
jmorra, line 1873: This would be a great place to link to the docs about the CLI tool you wrote for VW model creation.

src/site/markdown/model_formats.md
```diff
@@ -0,0 +1,1996 @@
...
+which might come from: 
+
+```bash
+echo "" | vw -f model.vw 2>/dev/null && cat model.vw | base64 && rm -f model.vw
+```
+
+##### (VW) vw modelUrl
+
+Exactly one of `model` or `modelUrl` is required.  When `modelUrl` is provided, it should be a string contain an
```
jmorra, line 1887: it should be a string containing

src/site/markdown/model_formats.md
```diff
@@ -0,0 +1,1996 @@
...
+[Apache VFS](https://commons.apache.org/proper/commons-vfs/filesystems.html) URL.  This is location from which the
```
jmorra: This is the location

jmorra:
I've added a few grammatical corrections.  After those are addressed I think it's ready to go.

Issue #37, opened 2015-08-28
Automatically add latest maven release to site


deaktator:
Use an async ajax request to maven central to automatically pull latest Aloha release.

jmorra:
This may seem a little late, but is this really necessary.  Is there ever a case where the current snapshot version is not x.x.x -1?

deaktator:
Yes.  deleted versions, etc.  You never know.  I think it's a bad
assumption.  Plus, if using semantic versioning, you might bump one of the
first two version numbers.

PR #38, opened 2015-08-28
Added javascript to make ajax call and updated the site.xml to includ…

deaktator:
…e it.  Works with 'mvn clean site && mvn -P stage site:stage


deaktator:
Pull request for #37

PR #48, opened 2015-09-10
Updating in_depth_walkthrough. added error handling stub.

jmorra:
Again a few more grammatical changes and I'll add notes in the PR.


src/site/markdown/error_handling.md
```diff
@@ -0,0 +1,4 @@
+# Error Handling
```
jmorra, line 1: What is the point of this file?  Is it linked from anywhere?  It also links to itself which I don't understand.

src/site/markdown/in_depth_walkthrough.md
```diff
@@ -0,0 +1,316 @@
...
+
+
+### Protobuf Plugin 
+
+Currently there are two main CompiledSemantics plugins.  The first is 
+[CompiledSemanticsProtoPlugin](aloha-core/scaladocs/index.html#com.eharmony.aloha.semantics.compiled.plugin.proto.CompiledSemanticsProtoPlugin) located 
+in the `com.eharmony.aloha.semantics.compiled.plugin.proto` package in [aloha-core](aloha-core/index.html). This 
+plugin allows for arbitrary kinds of [Protocol Buffer](https://developers.google.com/protocol-buffers/) instances 
+that extend [GeneratedMessage](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/GeneratedMessage).
+
```
jmorra, line 214: Can we add another sentence here about how protobufs are typesafe so they make integrating with Aloha really easy.  Also add something here about the protobuf versioning problems.  Finally add a plug for Thrift and Avro.  Something like "these are two other common communication formats that would be great for Aloha to support in the future, and it would be great if someone wants to contribute these to the project."

src/site/markdown/in_depth_walkthrough.md
```diff
@@ -0,0 +1,316 @@
...
+val oneCsvLine: CsvLineImpl = csvLines("72, 175") // height = 72 inches, weight = 175 lbs.
+val manyCsvLinesUsingVarArgs: Vector[CsvLineImpl] = csvLines("72, 175", "70, 180", "64, 110")
+val csvLineIterator: Iterator[CsvLineImpl] = csvLines(Iterator("72, 175", "70, 180", "64, 110"))
+val csvLineVector: Seq[CsvLineImpl] = csvLines(Seq("72, 175", "70, 180", "64, 110"))
+val nonStrictVectorOfCsvLine = csvLines.nonStrict(Vector("72, 175", "70, 180", "64, 110"))
+```
+
+## Factory
+
+Factories are important to create models:
```
jmorra, line 308: A few more sentences need to be added here.  At the very least a link to the wiki page about factory patterns.  Show an example with protobufs as well.  Also is this the appropriate place to show Java code for model instantiation as well?  What about code to create semantics?

src/site/markdown/in_depth_walkthrough.md
```diff
@@ -0,0 +1,316 @@
...
+val untypedFactory = ModelFactory.defaultFactory
+
+val typedFactory = untypedFactory.toTypedFactory[CsvLine, Double](semantics)
+
+val modelFile: java.io.File = ...
+
+// type: scala.util.Try[com.eharmony.aloha.models.Model[CsvLine, Double]]
+val modelAttempt = typedFactory.fromFile(modelFile)
+val model = modelAttempt.get 
+```
```
jmorra, line 338: Add a final section encouraging people to commit their own ideas as well for other formats and other model types.

src/site/markdown/in_depth_walkthrough.md
deaktator, line 214: I'll add this but this will require a substantial rewrite of a lot of Aloha.

jmorra, line 214: Of course, but the user should be aware of the versioning issues.  If they desire to use Aloha with Protobuf version /= 2.4.1 then they should be aware they have to recompile from source.

deaktator:
After mulling over these comments, I'm wondering about this page in general.  I see it as a more comprehensive _Getting Started_ doc.  I don't think it needs to be totally complete.  I should add additional pages that go into more depth about certain areas.  The best examples of this might be: 
1. Creating Factories
2. Creating Semantics (_CSV_, _proto_)
3. Error handling

Just about every issue could have side-by-side java / scala example code since the scala APIs are simpler.  Maybe even a _Spring_ page (or just a pointer to the spring unit test example).

jmorra:
This sounds like a great idea.

PR #54, opened 2015-10-26
closes #11

### LINKED ISSUE #11: Integrate Aloha with H2O
Add the ability to build H2O Aloha models.  This should have all the same support that VW currently has.

deaktator:
VW and H2O working.  I need some eyes on this one because I've been staring at it for a while and I think there may be some inelegant code in there.  Particularly, look at 
1. JSON input and code processing JSON input.  The parsing code is reminiscent of previous code that was not reused.
2. `ModelSource`.  Tell me what you think about this.
3.  Inner type members vs type parameters in ADTs.


aloha-core/src/main/scala/com/eharmony/aloha/io/sources/modelSources.scala
```diff
@@ -0,0 +1,114 @@
...
+   * Allows for the following forms.
+   * {{{
+   * { "model":    "[Base 64 encoded byte array here]" }
+   * { "modelUrl": "some://apache.vfs2/url/here" }
+   * { "modelUrl": "some://apache.vfs1/url/here", "via": "vfs1" }
+   * { "modelUrl": "some://apache.vfs2/url/here", "via": "vfs2" }
+   * { "modelUrl": "/absolute/path/here",         "via": "file" }
+   * }}}
+   */
+  implicit object jsonFormat extends RootJsonFormat[ModelSource] with DefaultJsonProtocol {
```
jmorra, line 31: I have remember reading (and I don't remember where) that it is usually good style to add return types to functions for readability.  What do you think?

aloha-core/src/main/scala/com/eharmony/aloha/io/sources/modelSources.scala
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  def shouldDelete = true
+
+  private[this] def copyToTmpFile(v: Vfs) = Try { v.fromByteArray(Base64.decodeBase64(b64EncodedData)) }
+
+  private[this] def getTmpVfs(tmpName: String) = Vfs.fromVfsType(vfsType)(tmpName)
+
+  private[this] def getTmpFileName = for {
+    sha1    <- Try { MessageDigest.getInstance("SHA-1") }
```
jmorra, line 104: This can be memoized

aloha-core/src/main/scala/com/eharmony/aloha/io/sources/modelSources.scala
```diff
@@ -0,0 +1,114 @@
...
+    localVfs.get // Will throw here if there was an issue.
+  }
+
+  def shouldDelete = true
+
+  private[this] def copyToTmpFile(v: Vfs) = Try { v.fromByteArray(Base64.decodeBase64(b64EncodedData)) }
+
+  private[this] def getTmpVfs(tmpName: String) = Vfs.fromVfsType(vfsType)(tmpName)
+
+  private[this] def getTmpFileName = for {
```
jmorra, line 103: This seems to have a lot of hardcoded values.  Do we want all of these hardcoded?  Also, do they all throw checked exceptions?

aloha-core/src/main/scala/com/eharmony/aloha/io/sources/modelSources.scala
```diff
@@ -0,0 +1,114 @@
...
+    sha1    <- Try { MessageDigest.getInstance("SHA-1") }
+    digest  <- Try { sha1.digest(b64EncodedData.getBytes) }
+    encoded <- Try { URLEncoder.encode(new String(digest), "UTF-8") }
+  } yield "tmp://" + encoded
```
jmorra, line 107: Can we get this from an enum in VFS?

aloha-core/src/main/scala/com/eharmony/aloha/io/sources/modelSources.scala
```diff
@@ -0,0 +1,114 @@
...
+
+final case class Base64StringSource(b64EncodedData: String, vfsType: VfsType = VfsType.vfs2) extends ModelSource {
+
+  /**
+   * Get a Vfs instance pointing to a local file containing the base64-decoded b64EncodedData.
+   * @return
+   */
+  def localVfs = {
+    val localVfs = for {
+      tmpName <- getTmpFileName
```
jmorra, line 88: Probably combine this line with the next, the `=` sign makes this a little confusing to read.

aloha-core/src/main/scala/com/eharmony/aloha/io/sources/modelSources.scala
```diff
@@ -0,0 +1,114 @@
...
+      v = getTmpVfs(tmpName)
+      _ <- copyToTmpFile(v)
```
jmorra, line 90: I might compress this into one function it'll make it more readable.

aloha-core/src/main/scala/com/eharmony/aloha/io/sources/modelSources.scala
```diff
@@ -0,0 +1,114 @@
...
+      local <- Try { v.replicatedToLocal() }
+    } yield local
+
+    localVfs.get // Will throw here if there was an issue.
+  }
+
+  def shouldDelete = true
+
+  private[this] def copyToTmpFile(v: Vfs) = Try { v.fromByteArray(Base64.decodeBase64(b64EncodedData)) }
```
jmorra, line 99: `fromByteArray` is poorly named as it actually writes and I wasn't aware of that until I read the function definition.  Maybe something like `writeToDisk`.

aloha-h2o/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+            <version>${h2o.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>log4j</groupId>
+                    <artifactId>log4j</artifactId>
+                </exclusion>
+
+                <!-- Use aloha's reflections:   org.reflections:reflections:0.9.9 -->
+                <!-- Hopefully this won't screw anything up, but internal forks are usually not frivolous. -->
+                <exclusion>
```
deaktator, line 62: I'm a little worried about this but it works under our tests.  This is somewhat disconcerting to exclude an internal fork of a library.

aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
```diff
@@ -0,0 +1,204 @@
...
+    featureNames: sci.IndexedSeq[String],
+    featureFunctions: sci.IndexedSeq[FeatureFunction[A]],
+    numMissingThreshold: Option[Int] = None)(implicit private[this] val scb: ScoreConverter[B])
+  extends BaseModel[A, B]
+     with Logging {
+
+  // Because H2o's RowData object is essentially a Map of String to Object, we unapply the wrapper
+  // and throw away the type information on the function return type.  We have type safety because
+  // FeatureFunction is sealed (ADT).
+  @transient private[this] lazy val lazyAnyRefFF = featureFunctions map {
```
deaktator, line 47: I hate having to throw away the typed information here but I don't know what else to do.

aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
```diff
@@ -0,0 +1,204 @@
...
+    failure(Seq(s"unknown categorical value ${e.getUnknownLevel} for variable: ${e.getColumnName}"), getMissingVariables(f.missing))
+
+  // TODO: Extract to trait.
+  protected[this] def getMissingVariables(missing: Map[String, Seq[String]]): Seq[String] =
+    missing.unzip._2.foldLeft(Set.empty[String])(_ ++ _).toIndexedSeq.sorted
+
+  protected[this] def failureDueToMissing(missing: Map[String, Seq[String]])(implicit audit: Boolean) =
+    failure(Seq(s"Too many features with missing variables: ${missing.count(_._2.nonEmpty)}"), getMissingVariables(missing))
+
+  protected[this] def constructFeatures(a: A): Features[RowData] = {
```
deaktator, line 91: immutable, no side effects as discussed.

PR #68, opened 2015-12-03
better vfs vw closes #67

### LINKED ISSUE #67: VW model creation forces VFS2
When creating an Aloha VW model and embedding the native model as a base64 encoded string VFS2 is forced.  This should not be the case and the format should be specified by the caller.

As part of this also check the H2O branch to make sure the same thing isn't happening over there.

jmorra:
This will now respect the user's choice of Vfs type when creating an Aloha VW JNI model.  This also fixes a bug where the via -> file type wasn't working correctly because the file name chosen was VFS style.

### DIFF
--- a/src/site/markdown/model_formats.md
+++ b/src/site/markdown/model_formats.md
@@ -1753,14 +1753,14 @@ needs to be trained on the same VW version that is included in the `aloha-vw-jni
     <td>N / A</td>
   </tr>
   <tr>
-    <td><a href="#aS_features">features</a></td>
+    <td><a href="#aVW_features">features</a></td>
     <td>Object</td>
     <td>N / A</td>
     <td>true</td>
     <td>N / A</td>
   </tr>
   <tr>
-    <td><a href="#aS_vw">vw</a></td>
+    <td><a href="#aVW_vw">vw</a></td>
     <td>Object</td>
     <td>N / A</td>
     <td>true</td>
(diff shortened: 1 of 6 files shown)


src/site/markdown/model_formats.md
```diff
@@ -1753,14 +1753,14 @@ needs to be trained on the same VW version that is included in the `aloha-vw-jni
     <td>N / A</td>
   </tr>
   <tr>
```
jmorra, line 1755: These links weren't working on the site page.  I'm not sure if this will fix the issue.  @deaktator please let me know if you think this will work.

deaktator:
I think that we need to get rid of the `via` field in `com.eharmony.aloha.io.sources.Base64StringSource`.  We can default to `com.eharmony.aloha.io.vfs.File`.  This will allow us to ditch conditionally reading and writing the local temp file using VFS 1 or VFS 2 depending on how the `via` field is populated.  

This doesn't seem like a huge benefit now, but it might in the future.  The reason is the the eH version of Pentaho's HDFS VFS plugin is 1.0.2 and the current version is 6.x.  In between those two versions, the plugin started using VFS 2.  Additionally, as of version 2.1 in VFS 2, HDFS appears to be natively supported.  

The issue is that currently, VFS 1 is used to read from HDFS and the models must explicitly state that VFS 1 is to be used.  If VFS 2 is used in some places and VFS 1 is used in others, an error occurs and this error will send the debugger on a fool's errand down the wrong path in an attempt to correct it.

So, it's necessary for the Aloha users to be unwaveringly consistent in their choice of VFS 1 or VFS 2.  But, using `com.eharmony.aloha.io.vfs.File` doesn't conflict with either so it may be used in conjunction with either choice.

deaktator:
I'm updating the code to reflect my above comment.

PR #78, opened 2016-01-19
Closes #77

### LINKED ISSUE #77: more descriptive cli error messages
I was working with Aloha and receiving the below error every now and then after altering the spec file. A more descriptive error message would be helpful. 

" Couldn't create writer for dataset type csv. Error: No applicable producer found.  Given Producer"

deaktator:


aloha-core/src/main/scala/com/eharmony/aloha/dataset/RowCreatorBuilder.scala
```diff
@@ -93,11 +93,12 @@ extends AlohaReadable[Try[B]]
     }
 
     private[this] def fail(failures: List[Failure[B]]): Failure[B] = {
-        info {
-            val msgs = producers.zip(failures).zipWithIndex.map{ case ((p, e), i) => s"$i)\t${p.name}: ${e.failed.get.getMessage}\n\t\t${e.failed.get.getStackTraceString.replaceAll("\n", "\n\t\t")}"}
-            msgs.mkString(s"${failures.size} failure${ if (failures.size == 1) "" else "s" } occurred while attempting to produce spec:\n\t", "\n\t", "")
-        }
-        Failure { new NoSuchElementException(s"No applicable producer found.  Given ${producers.map(_.name).mkString(", ")}") }
+      val s = if (failures.size == 1) "" else "s" // Pluralization
```
jmorra, line 96: Maybe this is overkill, but what do you think of some of the suggestions [here](http://stackoverflow.com/questions/5907296/java-api-for-plural-forms-of-english-words).

jmorra:
Code looks good.

jmorra:
This definitely now has the required information to fix the issue, but I'm still concerned about the format.  For someone who is super familiar with Java and the JVM the stacktrace is a natural thing to see, but for general purpose users I'm concerned that they'll have trouble interpreting it.  Is there anyway to put the cause as the first thing in the output and then some canned message about more info can be found below.  This is the singular most useful item and I think it should be highlighted.

PR #84, opened 2016-02-29
83 exploration library

jmorra:
This is the code to make use of [explore-java](https://github.com/multiworldtesting/explore-java) for Epsilon Greedy and Bootstrap models.  It should be notes that this won't compile now as the MWT code isn't in maven central.  As a result we need to decide upon how we want to handle this.

### DIFF
--- a/aloha-core/src/main/scala/com/eharmony/aloha/factory/ModelParser.scala
+++ b/aloha-core/src/main/scala/com/eharmony/aloha/factory/ModelParser.scala
@@ -47,6 +47,12 @@ trait ModelParser extends JsValuePimpz {
     final def getParser[A: RefInfo, B: RefInfo: JsonReader: ScoreConverter](factory: ModelFactory, semantics: Option[Semantics[A]]) = new Parser[JsValue, Model[A, B]] {
         def parse(json: JsValue): Model[A, B] = json.convertTo(modelJsonReader[A, B](factory, semantics))
     }
+
+    // This is a very slightly modified copy of the lift from Additional formats that removes the type bound.
+    protected[this] def lift[A](reader: JsonReader[A]) = new JsonFormat[A] {
+        def write(a: A): JsValue = throw new UnsupportedOperationException("No JsonWriter[" + a.getClass + "] available")
+        def read(value: JsValue) = reader.read(value)
+    }
 }
 
 trait BasicModelParser extends ModelParser {
--- a/aloha-core/src/main/scala/com/eharmony/aloha/factory/ModelParser.scala
+++ b/aloha-core/src/main/scala/com/eharmony/aloha/factory/ModelParser.scala
@@ -47,12 +47,6 @@ trait ModelParser extends JsValuePimpz {
     final def getParser[A: RefInfo, B: RefInfo: JsonReader: ScoreConverter](factory: ModelFactory, semantics: Option[Semantics[A]]) = new Parser[JsValue, Model[A, B]] {
         def parse(json: JsValue): Model[A, B] = json.convertTo(modelJsonReader[A, B](factory, semantics))
     }
-
-    // This is a very slightly modified copy of the lift from Additional formats that removes the type bound.
-    protected[this] def lift[A](reader: JsonReader[A]) = new JsonFormat[A] {
-        def write(a: A): JsValue = throw new UnsupportedOperationException("No JsonWriter[" + a.getClass + "] available")
-        def read(value: JsValue) = reader.read(value)
-    }
 }
 
 trait BasicModelParser extends ModelParser {
--- a/aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
+++ b/aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
@@ -4,7 +4,7 @@ import com.eharmony.aloha.factory.{ModelFactory, ModelParser, ParserProviderComp
 import com.eharmony.aloha.id.ModelIdentity
 import com.eharmony.aloha.models.{Model, BaseModel}
 import com.eharmony.aloha.score.Scores.Score
-import com.eharmony.aloha.score.basic.{ModelFailure, ModelOutput}
+import com.eharmony.aloha.score.basic.ModelOutput
 import com.eharmony.aloha.score.conversions.ScoreConverter
 import com.eharmony.aloha.score.conversions.ScoreConverter.Implicits.IntScoreConverter
 import com.eharmony.aloha.semantics.Semantics
@@ -42,25 +42,27 @@ case class BootstrapModel[A, B](
   override private[aloha] def getScore(a: A)(implicit audit: Boolean): (ModelOutput[B], Option[Score]) = {
     val mos = models.map(_.getScore(a))
 
-    // TODO This can be done better using ApplicativeFunctors and sequence from scalaz.
-    val scores = for {
-      m <- mos
-      result <- m._1.right.toOption
-    } yield result
+    // The basic idea here is that the type of mos is inverted, so the Either comes first and the results are all on
+    // the right.  If there is any failure then the Left will be evaluated, else all the successes will be combined into
+    // one IndexedSeq on the right.
+    val sequence = mos.foldLeft(Right(sci.IndexedSeq()):
+      Either[(ModelOutput[B], Option[Score]), sci.IndexedSeq[(Int, Option[Score])]]){ (a, b) =>
+      b match {
+        case (Left(error), os) => Left(failure(error._1, error._2, os))
+        case (Right(success), os) => a.right.map(_ :+ ((success, os)))
+      }
+    }
 
-    // Since we know we have at least 1 failure calling get on the option here is guaranteed to succeed.
-    // I know this is bad, the above to do should fix it.
-    if (scores.size != mos.size)
-      mos.collectFirst{ case (Left(error), os) => failure(error._1, error._2, os) }.get
-    else {
-      val decision = explorer.chooseAction(salt, scores)
-
-      val s = success(
-        score = classLabels(decision.getAction - 1),
-        subScores = mos.flatMap(_._2),
-        probability = Option(decision.getProbability)
-      )
-      s
+    sequence match {
+      case Left(error) => error
+      case Right(scores) =>
+        val decision = explorer.chooseAction(salt, scores.map(_._1))
+        val s = success(
+          score = classLabels(decision.getAction - 1),
+          subScores = scores.flatMap(_._2),
+          probability = Option(decision.getProbability)
+        )
+        s
     }
   }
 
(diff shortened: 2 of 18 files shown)


aloha-core/src/main/scala/com/eharmony/aloha/factory/ModelParser.scala
```diff
@@ -47,6 +47,12 @@ trait ModelParser extends JsValuePimpz {
     final def getParser[A: RefInfo, B: RefInfo: JsonReader: ScoreConverter](factory: ModelFactory, semantics: Option[Semantics[A]]) = new Parser[JsValue, Model[A, B]] {
         def parse(json: JsValue): Model[A, B] = json.convertTo(modelJsonReader[A, B](factory, semantics))
     }
+
+    // This is a very slightly modified copy of the lift from Additional formats that removes the type bound.
+    protected[this] def lift[A](reader: JsonReader[A]) = new JsonFormat[A] {
```
deaktator: This was extracted to the wrong place.  It should go in `com.eharmony.aloha.factory.ScalaJsonFormats` and should be implicit:

``` scala
implicit def lift[A](implicit reader: JsonReader[A]) = new JsonFormat[A] {
  def write(a: A): JsValue = throw new UnsupportedOperationException("No JsonWriter[" + a.getClass + "] available")
  def read(value: JsValue) = reader.read(value)
}
```

Then we could change the next few lines.

``` scala
import com.eharmony.aloha.factory.ScalaJsonFormats.lift // at top of file
val ast = json.convertTo[Ast[B]]
```

aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
```diff
@@ -0,0 +1,108 @@
...
+/**
+  * Created by jmorra on 2/26/16.
+  */
+case class BootstrapModel[A, B](
+  modelId: ModelIdentity,
+  models: sci.IndexedSeq[Model[A, Int]],
+  salt: GenAggFunc[A, Long],
+  classLabels: sci.IndexedSeq[B])(implicit scB: ScoreConverter[B]) extends BaseModel[A, B] {
+
+  @transient lazy val explorer = new BootstrapExplorer[sci.IndexedSeq[Int]](
```
deaktator: should be `private[this]` or `private[exploration]`

aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
```diff
@@ -0,0 +1,108 @@
...
+  override def chooseAction(actions: sci.IndexedSeq[Int]): Int = actions(index)
+}
+
+/**
+  * Created by jmorra on 2/26/16.
+  */
+case class BootstrapModel[A, B](
+  modelId: ModelIdentity,
+  models: sci.IndexedSeq[Model[A, Int]],
+  salt: GenAggFunc[A, Long],
```
deaktator, line 48: I would make a comment about what the salt function is for.  Just delete the comment and recreate it in IntelliJ and add a comment variable.

aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
```diff
@@ -0,0 +1,108 @@
...
+    * @param audit Whether the second field of the result Tuple2 should be Some (true) or None (false)
+    * @return a Tuple2 whose first field represents a simple version of the score, the second field (that should be
+    *         a Some instance if audit is true) is a more involved reporting of the score including errors and all
+    *         sub-model scores.
+    */
+  override private[aloha] def getScore(a: A)(implicit audit: Boolean): (ModelOutput[B], Option[Score]) = {
+    // The basic idea here is that we want a failure or a list of successes, so the Either comes first and the results
+    // are all on the right.  If there is any failure then the Left will be evaluated, else all the successes will be
+    // combined into one IndexedSeq on the right.
+    val mos = models.foldLeft(Right(sci.IndexedSeq()):
```
deaktator:
``` scala
models.foldLeft(Right(sci.IndexedSeq.empty)) ...
```

aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
```diff
@@ -0,0 +1,108 @@
...
+      Either[(ModelOutput[B], Option[Score]), sci.IndexedSeq[(Int, Option[Score])]]) { (x, y) =>
```
deaktator: This code seems like it's doing something you don't want it to do.  You are looping over the whole list of models and on failure throwing away the good scores.  But if a failure occurs, you keep going and throw away the previous failures.  I think that you should aggregate the failures or short circuit.  I do something like that in  `com.eharmony.aloha.util.EitherHelpers.mapSeq`.

deaktator: Seeing how this is used on lines 58, 61, I would make the right side a `Right [(sci.IndexedSeq[Int],  Option[sci.IndexedSeq[Score])]`.  Then you don't need to `map` / `flatMap` later.

aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
jmorra, line 48: Yea, I totally said I'd comment this very well and then didn't.  I'll add a bunch of comments.

aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
jmorra: Fixed

jmorra:
OK, now that explore-java is released all the checks pass, @deaktator please let me know when you've had a chance to review one more time.

aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
```diff
@@ -0,0 +1,138 @@
...
+  */
+private[this] case class NumberedPolicy(index: Int) extends Policy[sci.IndexedSeq[Int]] {
+  override def chooseAction(actions: sci.IndexedSeq[Int]): Int = actions(index)
+}
+
+/**
+  * A model for performing bootstrap style exploration.  This makes use of a number of policies.  The algorithm chooses
+  * one policy and then uses the other to calculate the appropriate probability of choosing that action.
+  * @param modelId a model identifier
+  * @param models a set of models that generate Int's.  These models MUST be deterministic for the probability to be correct.
```
deaktator, line 34: There should be a note in the comment for `models` that each model should return a value between 1 and `classLabels.size` inclusive.  I see it in `classLabels` but it is definitely better here.  The same goes for the epsilon-greeedy model.

aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/EpsilonGreedyModel.scala
```diff
@@ -0,0 +1,116 @@
...
+  override def chooseAction(action: Int): Int = action
+}
+
+/**
+  * A model which does epsilon greedy style exploration.  This will choose a random action with probability epsilon
+  * or an action from the defaultPolicy with probability 1 - epsilon.
+  * @param modelId a model identifier
+  * @param defaultPolicy the model to use for exploitation.  This MUST be deterministic for the probability to be correct.
+  *                      The model must return a value in the range 1 to `classLabels.size` (inclusive).
+  * @param epsilon the exploration/exploitation tradeoff parameter
```
deaktator: epsilon must be in the interval [0, 1]. 0 indicates never select an action randomly, 1 always select an action randomly.

aloha-core/src/test/scala/com/eharmony/aloha/models/exploration/EpsilonGreedyModelTest.scala
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  @Test def testClosed() {
+    val sub = new CloserTesterModel[Int]()
+    EpsilonGreedyModel(ModelId.empty, sub, 0.1f, GenFunc0("", (s: String) => 1l), sci.IndexedSeq(1, 2)).close()
+    assertTrue(sub.isClosed)
+  }
+
+  @Test def random() {
+    val epsilon = 0.9f
```
deaktator: epsilon should equal `1f`.

aloha-core/src/test/scala/com/eharmony/aloha/models/exploration/EpsilonGreedyModelTest.scala
```diff
@@ -0,0 +1,75 @@
...
+    val m = makeModel(1, epsilon, 1)
+    val s = m.getScore(null)
+    assertEquals(s._2.get.getScore.getProbability, epsilon / 3, delta)
+    assertEquals(s._1.right.get, "b")
```
deaktator: your assert statement parameters are reversed.

aloha-core/src/test/scala/com/eharmony/aloha/models/exploration/EpsilonGreedyModelTest.scala
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  @Test def policy() {
+    val epsilon = 0.1f
+    val m = makeModel(1, epsilon, 0)
+    val s = m.getScore(null)
+    assertEquals(s._2.get.getScore.getProbability, 1 - epsilon + epsilon / 3, delta)
```
deaktator: assert arguments are reversed.

aloha-core/src/test/scala/com/eharmony/aloha/models/exploration/EpsilonGreedyModelTest.scala
```diff
@@ -0,0 +1,75 @@
...
+  @Test def random() {
+    val epsilon = 0.9f
+    val m = makeModel(1, epsilon, 1)
+    val s = m.getScore(null)
+    assertEquals(s._2.get.getScore.getProbability, epsilon / 3, delta)
+    assertEquals(s._1.right.get, "b")
+  }
+
+  @Test def policy() {
+    val epsilon = 0.1f
```
deaktator: epsilon should equal 0.  And instead of directly checking that 

``` scala
s._1.right.get, "a"
```

you should create the RNG with the seed and check that `m.classLabels(RNG return value - 1)` equals `s._1.right.get`.  This way if you update MWT and they change the code, it might fail.  This would be good because then at least you would realize that something changed and that the model worked differently.

aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
jmorra, line 34: Noted.

PR #86, opened 2016-04-05
85 exploration sub scores

jmorra:
This will update sub score recording to only record those sub scores when using exploration type models if the sub model was actually used.  This is important because it allows evaluating the exploration model policies independently.

---

This change is [Reviewable](https://reviewable.io/reviews/eharmony/aloha/86)

### DIFF
--- a/aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
+++ b/aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
@@ -5,7 +5,6 @@ import com.eharmony.aloha.id.ModelIdentity
 import com.eharmony.aloha.models.{BaseModel, Model}
 import com.eharmony.aloha.score.Scores.Score
 import com.eharmony.aloha.score.basic.ModelOutput
-import com.eharmony.aloha.score.conversions.RelaxedConversions._
 import com.eharmony.aloha.score.conversions.ScoreConverter
 import com.eharmony.aloha.score.conversions.ScoreConverter.Implicits.IntScoreConverter
 import com.eharmony.aloha.semantics.Semantics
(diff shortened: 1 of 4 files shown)


aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala
```diff
@@ -78,9 +79,10 @@ case class BootstrapModel[A, B](
     successes: sci.IndexedSeq[Int] = sci.IndexedSeq.empty)(implicit audit: Boolean): (ModelOutput[B], Option[Score]) = {
     if (models.isEmpty) {
       val decision = explorer.chooseAction(salt(a), successes)
+      val usedSubScores = subScores.filter(_.getScore.getExtension(IntScore.impl).getScore == decision.getAction)
```
jmorra: I don't like this line at all, but I'm not sure of the best way to get either an Int or Long subscore, @deaktator, what do you think?

jmorra:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

_[aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala, line 82 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/86#-KEZ8VHlF0SGP4gOy3li-r1-82:-KEZY17z76mBQTmZwoUy:974905703) ([raw file](https://github.com/eharmony/aloha/blob/1172ed2826afb7d595d2a002d0bb748b98e9ca0e/aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/BootstrapModel.scala#L82)):_
I changed this to be a little more functional but I'm still not a fan of it.  I really want to use scalaz's sequence but I couldn't get that to work.  Ideally I would want a failure if any of the sub models fails to evaluate.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/86)_

jmorra:
Reviewed 2 of 4 files at r1, 2 of 2 files at r2.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/86)_

deaktator:
Reviewed 2 of 4 files at r1, 2 of 2 files at r2, 4 of 4 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

_[aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/EpsilonGreedyModel.scala, line 71 [r2]](https://reviewable.io:443/reviews/eharmony/aloha/86#-KEb5M7yF_HFECraeJw5:-KEb5M7yF_HFECraeJw6:-656918561) ([raw file](https://github.com/eharmony/aloha/blob/ea1810e6a045dc1a71d68674080bd8f7cbae5e66/aloha-core/src/main/scala/com/eharmony/aloha/models/exploration/EpsilonGreedyModel.scala#L71)):_
I like:

``` scala
subScores = os filter (_ => o == action)
```

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/86)_

jmorra:
Reviewed 1 of 4 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/86)_

jmorra:
Reviewed 1 of 4 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/86)_

PR #92, opened 2016-05-10
Adding new json file for use when data is in memory

jmorra:
This will allow users to create H2O models when the POJO and spec file are already in memory.

---

This change is [Reviewable](https://reviewable.io/reviews/eharmony/aloha/92)

### DIFF
--- a/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/Cli.scala
+++ b/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/Cli.scala
@@ -40,7 +40,7 @@ object Cli {
 
         val specVfs = Vfs.fromVfsType(vfsType)(spec)
         val modelVfs = Vfs.fromVfsType(vfsType)(model)
-        val json = H2oModel.json(specVfs, modelVfs, ModelId(id, name), externalModel, numMissingThresh, Option(notes))
+        val json = H2oModel.json(specVfs, modelVfs, ModelId(id, name), None, externalModel, numMissingThresh, Option(notes))
         println(json.compactPrint)
       case None => // Will be taken care of by scopt.
     }
--- a/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
+++ b/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
@@ -295,24 +295,10 @@ object H2oModel extends ParserProviderCompanion
     val notesList = notes filter {_.nonEmpty}
 
     features.map { fs =>
-      val updatedFs = responseColumn.fold(fs)(updateFeatureSpec(_, fs, "null"))
+      val updatedFs = responseColumn.fold(fs)(fs - _)
       val ast = H2oAst(H2oModel.parser.modelType, id, modelSource, updatedFs, numMissingThreshold, notesList)
       ast.toJson
     } getOrElse { throw new IllegalArgumentException(s"Couldn't get features from $spec.") }
-  }
-
-  // This will update a specific feature in the model while preserving the order.  Because of the order preservation
-  // this function is O(N).
-  private[this] def updateFeatureSpec(name: String, features: ListMap[String, H2oSpec], replacement: String): ListMap[String, H2oSpec] = {
-    features.map{ case(key, value) =>
-      if (key == name) {
-        value match {
-          case d: DoubleH2oSpec => (key, d.copy(spec = replacement))
-          case s: StringH2oSpec => (key, s.copy(spec = replacement))
-        }
-      }
-      else (key, value)
-    }
   }
 
   private[this] def getModelSource(model: Vfs, externalModel: Boolean): ModelSource =
--- a/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
+++ b/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
@@ -258,7 +258,8 @@ object H2oModel extends ParserProviderCompanion
     numMissingThreshold: Option[Int] = None,
     notes: Option[Seq[String]] = None): JsValue = {
     val modelSource = getModelSource(model, externalModel)
-    json(Right(spec), modelSource, id, responseColumn, numMissingThreshold, notes)
+    val features = getFeatures(spec)
+    json(spec.toString, features, modelSource, id, responseColumn, numMissingThreshold, notes)
   }
 
   @throws(classOf[IllegalArgumentException])
@@ -269,7 +270,8 @@ object H2oModel extends ParserProviderCompanion
     numMissingThreshold: Option[Int],
     notes: Option[Seq[String]]): JsValue = {
     val modelSource = getLocalSource(model.getBytes)
-    json(Left(spec.parseJson.asJsObject), modelSource, id, responseColumn, numMissingThreshold, notes)
+    val features = getFeatures(spec.parseJson.asJsObject)
+    json(spec, features, modelSource, id, responseColumn, numMissingThreshold, notes)
   }
 
   /**
@@ -283,14 +285,14 @@ object H2oModel extends ParserProviderCompanion
     */
   @throws(classOf[IllegalArgumentException])
   private[eharmony] def json(
-    spec: Either[JsObject, Vfs],
+    spec: String,
+    features: Option[ListMap[String, H2oSpec]],
     modelSource: ModelSource,
     id: ModelId,
     responseColumn: Option[String],
     numMissingThreshold: Option[Int],
     notes: Option[Seq[String]]): JsValue = {
     val notesList = notes filter {_.nonEmpty}
-    val features = spec.fold(getFeatures, getFeatures)
 
     features.map { fs =>
       val updatedFs = removeResponse(responseColumn, fs)
--- a/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
+++ b/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
@@ -21,7 +21,7 @@ import hex.genmodel.GenModel
 import hex.genmodel.easy.exception.PredictUnknownCategoricalLevelException
 import hex.genmodel.easy.{EasyPredictModelWrapper, RowData}
 import org.apache.commons.codec.binary.Base64
-import spray.json.{DeserializationException, JsArray, JsValue, JsonReader, pimpAny, pimpString}
+import spray.json._
 
 import scala.annotation.tailrec
 import scala.collection.immutable.ListMap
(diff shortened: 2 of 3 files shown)


deaktator:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 251 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/92#-KHRMu2fUiAe8xDN4i5Q:-KHRMu2fUiAe8xDN4i5R:1943257599) ([raw file](https://github.com/eharmony/aloha/blob/5c2bc3e5e028d890ad5b17a4400d0aa4e80511d8/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L251)):_

> ``` Scala
>     }
>   }
> ```

Do you need all of these?  I would just retain the one you want for model trainer and toss the others.  All of the implementations look fine to me.

The only thing I don't like is the `Either[JsObject, Vfs]`.  Why not just have two functions.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/92)_

deaktator:
Reviewed 1 of 1 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/92)_

jmorra:

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 251 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/92#-KHRMu2fUiAe8xDN4i5Q:-KHS9qfvfiSZvF8IdYsR:-663784087) ([raw file](https://github.com/eharmony/aloha/blob/5c2bc3e5e028d890ad5b17a4400d0aa4e80511d8/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L251)):_
<details><summary>

> Do you need all of these?  I would just retain the one you want for model trainer and toss the others.  All of the implementations look fine to me.
> 
> The only thing I don't like is the `Either[JsObject, Vfs]`.  Why not just have two functions.</details>
> This is just so we can display the `spec` in the error message.  Because of this we have to retain the `spec`. 

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/92)_

jmorra:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 251 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/92#-KHRMu2fUiAe8xDN4i5Q:-KHTOBrgB3AJMIfjGthI:307922835) ([raw file](https://github.com/eharmony/aloha/blob/5c2bc3e5e028d890ad5b17a4400d0aa4e80511d8/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L251)):_
<details><summary>

> This is just so we can display the `spec` in the error message.  Because of this we have to retain the `spec`. </details>
> I actually gone ahead and removed `Either`.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 309 [r2]](https://reviewable.io:443/reviews/eharmony/aloha/92#-KHTON79sbPie8gLTufZ:-KHTON79sbPie8gLTuf_:-120149910) ([raw file](https://github.com/eharmony/aloha/blob/7abfc6a62132b086b571873f2266263aa1aeb96d/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L309)):_

> ``` Scala
>       spec <- featureMap.get(r)
>     } yield {
>       val newSpec = spec match {
> ```

I'm really not happy with this as it won't generalize well.  I'm not sure how to do this in general as the trait `H2oSpec` doesn't have the copy function available on it.

---

_[aloha-h2o/src/test/scala/com/eharmony/aloha/models/h2o/H2oModelTest.scala, line 131 [r2]](https://reviewable.io:443/reviews/eharmony/aloha/92#-KHTO_5aUCVS6q-WmuA_:-KHTO_5aUCVS6q-WmuAa:-952694763) ([raw file](https://github.com/eharmony/aloha/blob/7abfc6a62132b086b571873f2266263aa1aeb96d/aloha-h2o/src/test/scala/com/eharmony/aloha/models/h2o/H2oModelTest.scala#L131)):_

> ``` Scala
>     val model = Vfs.fromVfsType(VfsType.vfs2)("res:com/eharmony/aloha/models/h2o/glm_afa04e31_17ad_4ca6_9bd1_8ab80005ce38.java")
>     val jsValue = H2oModel.json(spec, model, ModelId(1, "test-model"), Option("Sex"), true)
>     val sexField = jsValue.asJsObject.fields("features").asJsObject.fields("Sex")
> ```

Why are the quotes needed here?

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/92)_

deaktator:
Reviewed 2 of 3 files at r2, 1 of 1 files at r3.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 309 [r2]](https://reviewable.io:443/reviews/eharmony/aloha/92#-KHTON79sbPie8gLTufZ:-KHVHv_PHhLqEMPw0nc0:78683234) ([raw file](https://github.com/eharmony/aloha/blob/7abfc6a62132b086b571873f2266263aa1aeb96d/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L309)):_
<details><summary>

> I'm really not happy with this as it won't generalize well.  I'm not sure how to do this in general as the trait `H2oSpec` doesn't have the copy function available on it.</details>
> It's not general, but since `H2oSpec` is **sealed**, you'll get a compiler warning if you don't include all of the cases.  The real problem is the 

``` scala
featureMap + (r -> newSpec)
```

**THIS WILL NOT WORK**

``` scala
// Prints: "Map(1 -> one, 2 -> two, 4 -> four, 5 -> five, 3 -> th)"
println(
  ListMap(1 -> "one", 2 -> "two", 3 -> "three", 4 -> "four", 5 -> "five") + 
  (3 -> "th")
)
```

This will not replace the the key `3` _and maintain the original iteration order_.  This is important for the model to work correctly because when the underlying h2o model is built, it's underlying input vector expect the original iteration order.

At least I think this is the case.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/92)_

deaktator:
Reviewed 1 of 3 files at r2.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 309 [r2]](https://reviewable.io:443/reviews/eharmony/aloha/92#-KHTON79sbPie8gLTufZ:-KHVMDKpyYOyWwfnmK-G:-130831352) ([raw file](https://github.com/eharmony/aloha/blob/7abfc6a62132b086b571873f2266263aa1aeb96d/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L309)):_
<details><summary>

> It's not general, but since `H2oSpec` is **sealed**, you'll get a compiler warning if you don't include all of the cases.  The real problem is the 
> 
> ``` scala
> featureMap + (r -> newSpec)
> ```
> 
> **THIS WILL NOT WORK**
> 
> ``` scala
> // Prints: "Map(1 -> one, 2 -> two, 4 -> four, 5 -> five, 3 -> th)"
> println(
>   ListMap(1 -> "one", 2 -> "two", 3 -> "three", 4 -> "four", 5 -> "five") + 
>   (3 -> "th")
> )
> ```
> 
> This will not replace the the key `3` _and maintain the original iteration order_.  This is important for the model to work correctly because when the underlying h2o model is built, it's underlying input vector expect the original iteration order.
> 
> At least I think this is the case.</details>
> You'd have to do something like:

``` scala
responseColumn.fold(featureMap){c =>
  val (h, t) = featureMap.toVector.span{ case (k, v) => k != c }
  if (t.isEmpty) 
    featureMap
  else {
    val newSpec = t.head match {
      case d: DoubleH2oSpec => d.copy(spec = "null")
      case s: StringH2oSpec => s.copy(spec = "null")
    }
    ListMap((h ++ Seq(c -> newSpec) ++ h.drop(1)):_*)
  }
}
```

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/92)_

jmorra:

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 309 [r2]](https://reviewable.io:443/reviews/eharmony/aloha/92#-KHTON79sbPie8gLTufZ:-KHVUsxkpFv9LwNp9WDV:-688293113) ([raw file](https://github.com/eharmony/aloha/blob/7abfc6a62132b086b571873f2266263aa1aeb96d/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L309)):_
<details><summary>

> You'd have to do something like:
> 
> `````` scala
> responseColumn.fold(featureMap){c =>
>   val (h, t) = featureMap.toVector.span{ case (k, v) => k != c }
>   if (t.isEmpty) 
>     featureMap
>   else {
>     val newSpec = t.head match {
>       case d: DoubleH2oSpec => d.copy(spec = "null")
>       case s: StringH2oSpec => s.copy(spec = "null")
>     }
>     ListMap((h ++ Seq(c -> newSpec) ++ h.drop(1)):_*)
>   }
> }
> ```</details>
> I'm working on this now, but isn't this very dangerous in general as JSON specifically states that objects are unordered?
> ``````

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/92)_

jmorra:
Review status: 1 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 309 [r2]](https://reviewable.io:443/reviews/eharmony/aloha/92#-KHTON79sbPie8gLTufZ:-KHV_Mjop-c19OVMQbph:944694366) ([raw file](https://github.com/eharmony/aloha/blob/7abfc6a62132b086b571873f2266263aa1aeb96d/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L309)):_
<details><summary>

> I'm working on this now, but isn't this very dangerous in general as JSON specifically states that objects are unordered?</details>
> I couldn't seem to get your code to work properly.  I wrote it this way as it's simpler.  I know that it's an O(N) algorithm, but this is going to be called so infrequently that I'm ok with it if you are.

---

_[aloha-h2o/src/test/scala/com/eharmony/aloha/models/h2o/H2oModelTest.scala, line 135 [r4]](https://reviewable.io:443/reviews/eharmony/aloha/92#-KHV_DH2iwyGNrKMcq-x:-KHV_DH2iwyGNrKMcq-y:1501776459) ([raw file](https://github.com/eharmony/aloha/blob/7aa885f52e7f0f84533b54fbbcf4a5d26b08d462/aloha-h2o/src/test/scala/com/eharmony/aloha/models/h2o/H2oModelTest.scala#L135)):_

> ``` Scala
>     val jsValue = H2oModel.json(spec, model, ModelId(1, "test-model"), Option("Sex"), true)
>     val modelFeatures = jsValue.asJsObject.fields("features").asJsObject.fields
>     assertEquals("null", modelFeatures("Sex").asJsObject.fields("spec").toString.replaceAll("\"",""))
> ```

I added a test here to show order is preserved.  I still don't feel good about this as the type of `modelFeatures.keys` is a `Set`, hence there is no order guarantee.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/92)_

jmorra:
Review status: 1 of 3 files reviewed at latest revision, 5 unresolved discussions.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 298 [r5]](https://reviewable.io:443/reviews/eharmony/aloha/92#-KHX5i4937rWLHKoEH8E:-KHX5i4937rWLHKoEH8F:-639036580) ([raw file](https://github.com/eharmony/aloha/blob/77cfa81dd7dbe9d3234dc14229d34bab351f331b/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L298)):_

> ``` Scala
> 
>     features.map { fs =>
>       val updatedFs = responseColumn.fold(fs)(fs - _)
> ```

Per @deakator's comment since the features go through an internal map before being passed to the H2O pojo simply removing the feature here is sufficient.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/92)_

deaktator:
Reviewed 1 of 2 files at r5.
Review status: 2 of 3 files reviewed at latest revision, 5 unresolved discussions.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/92)_

deaktator:
Reviewed 1 of 2 files at r5.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/92)_

PR #97, opened 2016-05-14
Closes #96

### LINKED ISSUE #96: Remove H2O response column earlier
We have to remove the H2O response column before it's instantiated in the model to allow the response column to be a different type.

jmorra:
Got early removal of response column working.

---

This change is [Reviewable](https://reviewable.io/reviews/eharmony/aloha/97)

### DIFF
--- a/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
+++ b/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
@@ -315,9 +315,15 @@ object H2oModel extends ParserProviderCompanion
   private[this] def getFeatures(spec: JsObject, responseColumn: Option[String]): Option[ListMap[String, H2oSpec]] = {
     spec.getFields("features") match {
       case Seq(JsArray(fs)) =>
-        val features = fs.collect { case f if !responseColumn.exists(_ == f.asJsObject.fields("name").convertTo[String]) =>
+
+        // Note that the it is being assumed that an H2oSpec cannot be instantiated for the response column
+        // hence we cannot call f.convertTo[H2oSpec] on the response column.
+        def convertToSpec(f: JsValue) = {
           val s = f.convertTo[H2oSpec]
           (s.name, s)
+        }
+        val features = responseColumn.fold(fs.map(convertToSpec)){ r =>
+          fs.collect { case f if r != f.asJsObject.fields("name").convertTo[String] => convertToSpec(f)}
         }
         Some(ListMap(features:_*))
       case _ => None
--- a/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
+++ b/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
@@ -258,8 +258,8 @@ object H2oModel extends ParserProviderCompanion
     numMissingThreshold: Option[Int] = None,
     notes: Option[Seq[String]] = None): JsValue = {
     val modelSource = getModelSource(model, externalModel)
-    val features = getFeatures(spec)
-    json(spec.toString, features, modelSource, id, responseColumn, numMissingThreshold, notes)
+    val features = getFeatures(spec, responseColumn)
+    json(spec.toString, features, modelSource, id, numMissingThreshold, notes)
   }
 
   @throws(classOf[IllegalArgumentException])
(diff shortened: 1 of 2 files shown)


jmorra:
It should be noted that this is necessary so that we can have a response column type of boolean so we can train a model which targets a boolean learner and not only a regression type learner.

<details><summary>

> #### Closes #96
> 
> Got early removal of response column working.
> 
> </details>

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 316 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/97#-KHlAgWG8hA81h1ChxK8:-KHlAgWG8hA81h1ChxK9:977138674) ([raw file](https://github.com/eharmony/aloha/blob/5adfd9c85e6b83352b61857e16950f76d035c5a7/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L316)):_

> ``` Scala
>   private[this] def getFeatures(spec: JsObject, responseColumn: Option[String]): Option[ListMap[String, H2oSpec]] = {
>     spec.getFields("features") match {
>       case Seq(JsArray(fs)) =>
> ```

I'm not a big fan of how this works.  There has to be an easier way to get the name field, especially without the quote removal.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/97#-:-KHlApAzKpidAFku4MT0:-903028978)_

jmorra:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 316 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/97#-KHlAgWG8hA81h1ChxK8:-KHvKTs1ZhptTn1AeJQv:-414479429) ([raw file](https://github.com/eharmony/aloha/blob/5adfd9c85e6b83352b61857e16950f76d035c5a7/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L316)):_
<details><summary>

> I'm not a big fan of how this works.  There has to be an easier way to get the name field, especially without the quote removal.</details>
> This seems a lot better now, Ryan, what do you think?

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/97)_

deaktator:
Reviewed 2 of 2 files at r2.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 316 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/97#-KHlAgWG8hA81h1ChxK8:-KHzZxYzRUNqj94BzOE-:160425391) ([raw file](https://github.com/eharmony/aloha/blob/5adfd9c85e6b83352b61857e16950f76d035c5a7/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L316)):_
<details><summary>

> This seems a lot better now, Ryan, what do you think?</details>
> yup

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/categories/package.scala, line 26 [r2]](https://reviewable.io:443/reviews/eharmony/aloha/97#-KHz_JW-aof7RqUbk15Q:-KHz_JW-aof7RqUbk15R:176668187) ([raw file](https://github.com/eharmony/aloha/blob/e97fa7175b77c4c82ce16f2a0e3c343458341573/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/categories/package.scala#L26)):_

> ``` Scala
> 
>   final def predictor[B: RefInfo](model: EasyPredictModelWrapper): Either[TypeCoercionNotFound, RowData => Either[IllConditioned, B]] =
>     TypeCoercion[C, B](ri, RefInfo[B]) map { coercion =>
> ```

I like this a lot better.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/categories/package.scala, line 78 [r2]](https://reviewable.io:443/reviews/eharmony/aloha/97#-KHzYXqIcnFfzDm8oUwp:-KHzYXqIcnFfzDm8oUwq:464513830) ([raw file](https://github.com/eharmony/aloha/blob/e97fa7175b77c4c82ce16f2a0e3c343458341573/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/categories/package.scala#L78)):_

> ``` Scala
>   // classProbabilities is an Array of size 2.  We are assuming that a binomial model has a positive and negative class.
>   // We are also assuming that the label for the negative class comes lexographically before the label for the positive
>   // class.  If these assumptions are met then classProbabilities(1) will be the probability that the given example
> ```

I would add a little note that this is supported in several common cases for string representations of binary response variables:

``` scala
"0" <= "1"         // true
"-1" <= "1"        // true
"neg" <= "pos"     // true
"false" <= "true"  // true
"False" <= "True"  // true
```

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/97)_

PR #100, opened 2016-05-20
99 h2o serialization

jmorra:
This change now constructs the H2O model with the `GenModel` already made.  This is advantageous because when moving the models between machines it makes use of the same `classCacheDir` that is used for the Aloha feature functions.  This means that if that directory is uploaded to the cluster than the whole H2O model will work correctly.

---

This change is [Reviewable](https://reviewable.io/reviews/eharmony/aloha/100)

### DIFF
--- a/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
+++ b/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
@@ -36,12 +36,12 @@ import scala.util.{Failure, Success, Try}
 /**
  * Created by deak on 9/30/15.
  */
-final case class H2oModel[-A, +B](
+final case class H2oModel[-A, +B : ScoreConverter](
     modelId: ModelIdentity,
     h2OModel: GenModel,
     featureNames: sci.IndexedSeq[String],
     featureFunctions: sci.IndexedSeq[FeatureFunction[A]],
-    numMissingThreshold: Option[Int] = None)(implicit private[this] val scb: ScoreConverter[B], implicit private[this] val rib: RefInfo[B])
+    numMissingThreshold: Option[Int] = None)
   extends BaseModel[A, B]
      with Logging {
 
@@ -54,19 +54,13 @@ final case class H2oModel[-A, +B](
   }
 
   @transient private[this] lazy val h2OPredictor: (RowData) => Either[IllConditioned, B] = {
-    val retrieval = H2oModelCategory.predictor[B](new EasyPredictModelWrapper(h2OModel))
-    mapRetrievalError[B](retrieval).get
+    val retrieval = H2oModelCategory.predictor[B](new EasyPredictModelWrapper(h2OModel))(implicitly[ScoreConverter[B]].ri)
+    H2oModel.mapRetrievalError[B](retrieval)(implicitly[ScoreConverter[B]].ri).get
   }
 
   // Force initialization of lazy vals.
   require(lazyAnyRefFF != null)
   require(h2OPredictor != null)
-
-  protected[h2o] def mapRetrievalError[B: RefInfo](retrieval: Either[PredictionFuncRetrievalError, RowData => Either[IllConditioned, B]]) = retrieval match {
-    case Right(f) => Success(f)
-    case Left(UnsupportedModelCategory(category)) => Failure(new UnsupportedOperationException(s"In model ${h2OModel.getClass.getCanonicalName}: ModelCategory ${category.name} non supported."))
-    case Left(TypeCoercionNotFound(category)) => Failure(new IllegalArgumentException(s"In model ${h2OModel.getClass.getCanonicalName}: Could not ${category.name} model to Aloha output type: ${RefInfoOps.toString[B]}."))
-  }
 
   override private[aloha] def getScore(a: A)(implicit audit: Boolean): (ModelOutput[B], Option[Score]) = {
     val f = constructFeatures(a)
@@ -221,7 +215,6 @@ object H2oModel extends ParserProviderCompanion
           case Right(featureMap) =>
             val (names, functions) = featureMap.toIndexedSeq.unzip
             val genModel = getGenModelFromFile(h2o.modelSource, classCacheDir)
-            implicit val rib = scB.ri
             H2oModel[A, B](h2o.modelId,
               genModel,
               names,
@@ -248,6 +241,12 @@ object H2oModel extends ParserProviderCompanion
       }
       case _ => None
     }
+  }
+
+  protected[h2o] def mapRetrievalError[B: RefInfo](retrieval: Either[PredictionFuncRetrievalError, RowData => Either[IllConditioned, B]]) = retrieval match {
+    case Right(f) => Success(f)
+    case Left(UnsupportedModelCategory(category)) => Failure(new UnsupportedOperationException(s"In model ${classOf[H2oModel[_, _]].getCanonicalName}: ModelCategory ${category.name} non supported."))
+    case Left(TypeCoercionNotFound(category)) => Failure(new IllegalArgumentException(s"In model ${classOf[H2oModel[_, _]].getCanonicalName}: Could not ${category.name} model to Aloha output type: ${RefInfoOps.toString[B]}."))
   }
 
   protected[h2o] def getGenModel[B, C](
--- a/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
+++ b/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
@@ -45,7 +45,7 @@ final case class H2oModel[-A, +B](
   extends BaseModel[A, B]
      with Logging {
 
-  // Because H2o's RowData object is essentially a Map of String to Object, we unapply the wrapper
+  // Because H2O's RowData object is essentially a Map of String to Object, we unapply the wrapper
   // and throw away the type information on the function return type.  We have type safety because
   // FeatureFunction is sealed (ADT).
   @transient private[this] lazy val lazyAnyRefFF = featureFunctions map {
--- a/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
+++ b/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
@@ -1,5 +1,7 @@
 package com.eharmony.aloha.models.h2o
 
+import java.io.File
+
 import com.eharmony.aloha.factory.{ModelParser, ModelParserWithSemantics, ParserProviderCompanion}
 import com.eharmony.aloha.id.{ModelId, ModelIdentity}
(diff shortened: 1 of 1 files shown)


jmorra:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 51 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/100#-KIEs4sdA93qRju863af:-KIEs4sdA93qRju863ag:-1156905393) ([raw file](https://github.com/eharmony/aloha/blob/1c61050917d4e56e8478db09cd6d4dab537a0f81/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L51)):_

> ``` Scala
>   // and throw away the type information on the function return type.  We have type safety because
>   // FeatureFunction is sealed (ADT).
>   private[this] val anyRefFF = featureFunctions map {
> ```

I'm not sure if this should stay as a `lazy val` or not.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 153 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/100#-KIEsAFWcxwotOZ2BsR-:-KIEsAFWcxwotOZ2BsR0:-1342100416) ([raw file](https://github.com/eharmony/aloha/blob/1c61050917d4e56e8478db09cd6d4dab537a0f81/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L153)):_

> ``` Scala
>       }
> 
>     features(0, anyRefFF.size, new RowData, Map.empty, anyRefFF)
> ```

If we want to make  anyRefFF lazy again, this needs to go back to the way it was.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 225 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/100#-KIEsMKasdJB_WiyQUM7:-KIEsMKasdJB_WiyQUM8:1074497804) ([raw file](https://github.com/eharmony/aloha/blob/1c61050917d4e56e8478db09cd6d4dab537a0f81/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L225)):_

> ``` Scala
>   }
> 
>   protected[h2o] def getClassCacheDir[A](semantics: Semantics[A]): Option[File] = {
> ```

I'm not a huge fan of this function.  While it works I think it could be a little cleaner with a `for` comprehension.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/100)_

deaktator:
Looks fine.  I would wrap the `RowData => Either[IllConditioned, B]` in a wrapper class that takes the `GenModel` and extends `RowData => Either[IllConditioned, B]`.

Not sure if this change breaks the temp file / dir deletion.

<details><summary>

> #### 99 h2o serialization
> 
> This change now constructs the H2O model with the `GenModel` already made.  This is advantageous because when moving the models between machines it makes use of the same `classCacheDir` that is used for the Aloha feature functions.  This means that if that directory is uploaded to the cluster than the whole H2O model will work correctly.
> 
> </details>

---

Reviewed 1 of 1 files at r1.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 41 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/100#-KIF4ij6QawNuBEsMVNC:-KIF4ij70BeGjym1_aMY:1354677487) ([raw file](https://github.com/eharmony/aloha/blob/1c61050917d4e56e8478db09cd6d4dab537a0f81/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L41)):_

> ``` Scala
> final case class H2oModel[-A, +B](
>     modelId: ModelIdentity,
>     h2oPredictor: RowData => Either[IllConditioned, B],
> ```

I think that my biggest complaint to the whole change set is that this is a raw function.  I think that it if is exposed for the _"user"_ to provide, it should be something like `H2oGenModel[B]`, where `H2oGenModel` is defined as: 

``` scala
case class H2oGenModel[+A](model: GenModel) extends (RowData => Either[IllConditioned, A]) {
  def apply(x: RowData):  Either[IllConditioned, A] = { 
    val y = model.predict(x)
    // TODO: Transform y to a value of type A.
    ??? 
  }
}
```

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 51 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/100#-KIEs4sdA93qRju863af:-KIF6B5fEtAbL7O6MQmF:1857699590) ([raw file](https://github.com/eharmony/aloha/blob/1c61050917d4e56e8478db09cd6d4dab537a0f81/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L51)):_
<details><summary>

> I'm not sure if this should stay as a `lazy val` or not.
> </details>
> I would suggest one of two things: 
1. Leave this `@transient lazy val` - this was originally made a `@transient lazy val` not because it doesn't Serialize, but because it's unnecessary to do so since it's a direct mapping from `featureFunctions`.  Since the `lazy val` is dereference in the prediction function just once, there is a nominal (_if any_) performance difference.  
2. Encode `featureFunctions` as an unboxed union codomain via Miles Sabin's technique using the  Curry-Howard isomorphism.<sup>1</sup>

(2) would look like this: 

``` scala
case class H2oModel[..., T: (jl.Double |∨| String)#λ ](
...
sci.IndexedSeq[A => T]
```
1. [Unboxed union types in Scala via the Curry-Howard isomorphism](http://milessabin.com/blog/2011/06/09/scala-union-types-curry-howard/)

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 225 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/100#-KIEsMKasdJB_WiyQUM7:-KIFB14VlMofHovLFwDs:-207672209) ([raw file](https://github.com/eharmony/aloha/blob/1c61050917d4e56e8478db09cd6d4dab537a0f81/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L225)):_
<details><summary>

> I'm not a huge fan of this function.  While it works I think it could be a little cleaner with a `for` comprehension.
> </details>
> You could do something like: 

``` scala
Option(semantics) collect { 
  case c: CompiledSemantics[A] => c.compiler 
} flatMap { 
  Option(_) collect { 
    case t: TwitterEvalCompiler => t.classCacheDir 
  } 
}  
```

_OR_

``` scala
for {
  s <- Option(semantics) filter {_.isInstanceOf[CompiledSemantics[A]]}
  c <- Option(s.compiler) filter {_.isInstanceOf[TwitterEvalCompiler]}
} yield c.classCacheDir
```

but that just seems super _clever_ for no reason.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 255 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/100#-KIFCbfZwFSXAUmsMggs:-KIFCbfZwFSXAUmsMggt:1739317778) ([raw file](https://github.com/eharmony/aloha/blob/1c61050917d4e56e8478db09cd6d4dab537a0f81/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L255)):_

> ``` Scala
>   }
> 
>   private[this] def getPredictor[B](
> ```

I would have this return `H2oGenModel` like I mentioned in the comment near the top of the review.1

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/100#-:-KIFCwtoCqn1FAdYD6Ix:-221405548)_

jmorra:
Review status: 0 of 1 files reviewed at latest revision, 4 unresolved discussions.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 41 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/100#-KIF4ij6QawNuBEsMVNC:-KIThD0Vo3DrTzusEGwB:502427610) ([raw file](https://github.com/eharmony/aloha/blob/1c61050917d4e56e8478db09cd6d4dab537a0f81/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L41)):_
<details><summary>

> I think that my biggest complaint to the whole change set is that this is a raw function.  I think that it if is exposed for the _"user"_ to provide, it should be something like `H2oGenModel[B]`, where `H2oGenModel` is defined as: 
> 
> ``` scala
> case class H2oGenModel[+A](model: GenModel) extends (RowData => Either[IllConditioned, A]) {
>   def apply(x: RowData):  Either[IllConditioned, A] = { 
>     val y = model.predict(x)
>     // TODO: Transform y to a value of type A.
>     ??? 
>   }
> }
> ```
> 
> </details>
> I'm not sure exactly what you mean by using a case class here, but I've made the model easier to instantiate by accepting the H2O `GenModel` directly.  Let me know if this is better to you.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 255 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/100#-KIFCbfZwFSXAUmsMggs:-KIThR9fQz5BxKfnxETB:676635175) ([raw file](https://github.com/eharmony/aloha/blob/1c61050917d4e56e8478db09cd6d4dab537a0f81/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L255)):_
<details><summary>

> I would have this return `H2oGenModel` like I mentioned in the comment near the top of the review.1
> </details>
> Done

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/100)_

deaktator:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 41 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/100#-KIF4ij6QawNuBEsMVNC:-KIYdyn6oZwvaGLJGggV:-89963163) ([raw file](https://github.com/eharmony/aloha/blob/1c61050917d4e56e8478db09cd6d4dab537a0f81/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L41)):_
<details><summary>

> I'm not sure exactly what you mean by using a case class here, but I've made the model easier to instantiate by accepting the H2O `GenModel` directly.  Let me know if this is better to you.
> </details>
> This is better than what I was thinking.  My suggestion was somewhat myopic I think.  I was thinking that all of the logic to produce a `Either[IllConditioned, B]` would go in here.  That turns out to be super dumb because all of that logic is what makes the `H2oModel` an `H2oModel`.  

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/100)_

PR #108, opened 2016-08-17
Providing fix for compiling H2O models with custom classloader

jmorra:
Compiling H2O models within Jetty proved unsuccessful.  The reason for this is that Jetty has its own classloader.  A workaround for this is to find the `h2o-genmodel` jar within the current Thread's classloader.  While this will fix the specific need of compiling H2O models within Jetty, it should also provide a more general framework for dealing with custom classloaders.

---

This change is [Reviewable](https://reviewable.io/reviews/eharmony/aloha/108)

### DIFF
--- a/aloha-h2o/pom.xml
+++ b/aloha-h2o/pom.xml
@@ -24,6 +24,7 @@
 
     <properties>
         <h2o.version>3.8.2.3</h2o.version>
+        <h2o.genmodel.artifact>h2o-genmodel</h2o.genmodel.artifact>
     </properties>
 
     <dependencies>
@@ -59,7 +60,7 @@
         </dependency>
         <dependency>
             <groupId>ai.h2o</groupId>
-            <artifactId>h2o-genmodel</artifactId>
+            <artifactId>${h2o.genmodel.artifact}</artifactId>
             <version>${h2o.version}</version>
             <exclusions>
                 <exclusion>
@@ -103,6 +104,26 @@
                 <artifactId>maven-resources-plugin</artifactId>
                 <version>2.3</version>
                 <executions>
+                    <execution>
+                        <id>copy-resources</id>
+                        <phase>process-resources</phase>
+                        <goals>
+                            <goal>copy-resources</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>target/classes</outputDirectory>
+                            <encoding>UTF-8</encoding>
+                            <resources>
+                                <resource>
+                                    <directory>src/main/resources</directory>
+                                    <filtering>true</filtering>
+                                    <includes>
+                                        <include>h2o_mvn.properties</include>
+                                    </includes>
+                                </resource>
+                            </resources>
+                        </configuration>
+                    </execution>
                     <execution>
                         <id>copy-test-resources</id>
                         <phase>process-test-resources</phase>
--- a/aloha-h2o/src/main/resources/h2o_mvn.properties
+++ b/aloha-h2o/src/main/resources/h2o_mvn.properties
@@ -1,0 +1,1 @@
+h2oGenModelName=${h2o.genmodel.artifact}-${h2o.version}.jar
--- a/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
+++ b/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala
@@ -251,12 +251,20 @@ object H2oModel extends ParserProviderCompanion
     case Left(TypeCoercionNotFound(category)) => Failure(new IllegalArgumentException(s"In model ${classOf[H2oModel[_, _]].getCanonicalName}: Could not ${category.name} model to Aloha output type: ${RefInfoOps.toString[B]}."))
   }
 
-  private[h2o] def getJar(collectFn: URL => Boolean): Array[String] =
-    Compiler.currentClassLoader match {
-      case urlClassLoader: URLClassLoader => urlClassLoader.getURLs.collect{case x if collectFn(x) => x.toString}
-      case _ => Array.empty[String]
+  private[h2o] def getJar(collectFn: URL => Boolean): Array[File] =
+    currentClassLoader match {
+      case urlClassLoader: URLClassLoader => urlClassLoader.getURLs.flatMap { url =>
+        // The File constructor can throw if the URL does not reference a local file.  This might be possible if
+        // running in some kind of applet.
+        if (collectFn(url)) Try(new File(url.toURI)).toOption
+        else None
+      }
+      case _ => Array.empty[File]
     }
 
+  // This guarantees that the h2oGenModelName property used below is guaranteed to be in sync with the maven artifact
+  // dependency defined in the POM. This is because the h2o_mvn.properties is a filtered resource, meaning maven injects
+  // values from the build into the properties file at build time.
   private[h2o] lazy val h2oProps = {
     val stream = getClass.getClassLoader.getResourceAsStream("h2o_mvn.properties")
     try {
@@ -266,6 +274,8 @@ object H2oModel extends ParserProviderCompanion
     }
     finally stream.close()
   }
+
+  private[this] lazy val currentClassLoader = Thread.currentThread().getContextClassLoader
 
   protected[h2o] def getGenModel[B, C](
     input: => C,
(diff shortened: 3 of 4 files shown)


gpetrosk:
Tested within scorer-service lt environment and it works. This is still pending a test from scorer

jmorra:
Tested within our training environment within Spark and it works fine.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/108#-:-KPTfGCQO6ixKkwHZpz-:bc48lbi)_

deaktator:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

_[aloha-h2o/src/main/resources/h2o_mvn.properties, line 1 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/108#-KPTu1MYPaDZncD-pXo-:-KPTu1MYPaDZncD-pXo0:b-ffxoy3) ([raw file](https://github.com/eharmony/aloha/blob/a2ea2e485ea66bb8d9371e20a364ea30559037f1/aloha-h2o/src/main/resources/h2o_mvn.properties#L1)):_

> ``` INI
> h2oGenModelName=${h2o.genmodel.artifact}-${h2o.version}.jar
> ```

_Beautiful!_

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 260 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/108#-KPTszqOAH23gkqotCNr:-KPTszqOAH23gkqotCNs:b-oa9ws4) ([raw file](https://github.com/eharmony/aloha/blob/a2ea2e485ea66bb8d9371e20a364ea30559037f1/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L260)):_

> ``` Scala
>     }
> 
>   private[h2o] lazy val h2oProps = {
> ```

I would put some scaladoc or plain old comments here explaining why this is done.  It really is a great idea and whoever came up with this should probably get a raise but the motivation is likely to be lost on any reader who wasn't around for the discussion as to why this rocks.  

For those not in the know, this guarantees that the `h2oGenModelName` property used below is guaranteed to be in sync with the maven artifact dependency defined in the POM.  This is because the `h2o_mvn.properties` is a filtered resource, meaning maven injects values from the build into the properties file at build time.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/compiler/Compiler.scala, line 30 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/108#-KPTvPbuwzieQII6MoEt:-KPTvPbuwzieQII6MoEu:bap8xhl) ([raw file](https://github.com/eharmony/aloha/blob/a2ea2e485ea66bb8d9371e20a364ea30559037f1/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/compiler/Compiler.scala#L30)):_

> ``` Scala
>  * @tparam B the output type.  This should be an interface or base class that the compiled class extends.
>  */
> private[h2o] class Compiler[B](compilationClassPaths: Iterable[String] = Iterable.empty,
> ```

Thinking about this, I think `compilationClassPaths` should be an `Iterable[File]` since it is really what we mean by `String` these are file paths that are concatenated to make an optional classpath argument that will eventually be passed to the java compiler.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/compiler/Compiler.scala, line 141 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/108#-KPU0--f71r84B2d8g5f:-KPU0--f71r84B2d8g5g:b-4q0ik0) ([raw file](https://github.com/eharmony/aloha/blob/a2ea2e485ea66bb8d9371e20a364ea30559037f1/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/compiler/Compiler.scala#L141)):_

> ``` Scala
> 
> private[h2o] object Compiler extends Logging {
>   def currentClassLoader = Thread.currentThread().getContextClassLoader
> ```

I think this one line is the most contentious part of the entire change set.  I just don't know what the right thing is here.  I'm worried about issues like `java.lang.Thread` has a `setContextClassLoader` method.  Also, `currentThread` will refer to different threads at different times.  Can different threads have different context class loaders?  I assume so.  Therefore, it seems to me that this should only be called once per model instantiation or possibly even once per JVM instance.

Maybe what that means is that this should moved into a lazy value inside the `H2oModel` instance and then passed around?  I don't know.  I just really worry about this (probably way too much).

It seems to me that the construction of the ClassLoader in the `instantiate` function maybe doesn't need access to the host ClassLoader.  Instead of:

``` scala
val classLoader = new URLClassLoader(Array(compileDir.toURI.toURL), Compiler.currentClassLoader)
```

Maybe just

``` scala
val classLoader = new URLClassLoader(Array(compileDir.toURI.toURL) ++ compilationClassPaths)
```

and the `compilationClassPaths` constructor parameter should be an `Iterable[URL]`.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/108)_

deaktator:
Reviewed 4 of 4 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/108)_

jmorra:
Review status: 2 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala, line 260 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/108#-KPTszqOAH23gkqotCNr:-KPUGZ3bFjWzlMMX43AI:bb6un1j) ([raw file](https://github.com/eharmony/aloha/blob/a2ea2e485ea66bb8d9371e20a364ea30559037f1/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/H2oModel.scala#L260)):_
<details><summary>

> I would put some scaladoc or plain old comments here explaining why this is done.  It really is a great idea and whoever came up with this should probably get a raise but the motivation is likely to be lost on any reader who wasn't around for the discussion as to why this rocks.  
> 
> For those not in the know, this guarantees that the `h2oGenModelName` property used below is guaranteed to be in sync with the maven artifact dependency defined in the POM.  This is because the `h2o_mvn.properties` is a filtered resource, meaning maven injects values from the build into the properties file at build time.
> </details>
> Done

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/compiler/Compiler.scala, line 30 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/108#-KPTvPbuwzieQII6MoEt:-KPUGX0aDsM3If3qesv0:b-8ayu6x) ([raw file](https://github.com/eharmony/aloha/blob/a2ea2e485ea66bb8d9371e20a364ea30559037f1/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/compiler/Compiler.scala#L30)):_
<details><summary>

> Thinking about this, I think `compilationClassPaths` should be an `Iterable[File]` since it is really what we mean by `String` these are file paths that are concatenated to make an optional classpath argument that will eventually be passed to the java compiler.
> </details>
> Agree

---

_[aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/compiler/Compiler.scala, line 141 [r1]](https://reviewable.io:443/reviews/eharmony/aloha/108#-KPU0--f71r84B2d8g5f:-KPUG51G0kWt9zI3ubo9:bhmzx1h) ([raw file](https://github.com/eharmony/aloha/blob/a2ea2e485ea66bb8d9371e20a364ea30559037f1/aloha-h2o/src/main/scala/com/eharmony/aloha/models/h2o/compiler/Compiler.scala#L141)):_
<details><summary>

> I think this one line is the most contentious part of the entire change set.  I just don't know what the right thing is here.  I'm worried about issues like `java.lang.Thread` has a `setContextClassLoader` method.  Also, `currentThread` will refer to different threads at different times.  Can different threads have different context class loaders?  I assume so.  Therefore, it seems to me that this should only be called once per model instantiation or possibly even once per JVM instance.
> 
> Maybe what that means is that this should moved into a lazy value inside the `H2oModel` instance and then passed around?  I don't know.  I just really worry about this (probably way too much).
> 
> It seems to me that the construction of the ClassLoader in the `instantiate` function maybe doesn't need access to the host ClassLoader.  Instead of:
> 
> ``` scala
> val classLoader = new URLClassLoader(Array(compileDir.toURI.toURL), Compiler.currentClassLoader)
> ```
> 
> Maybe just
> 
> ``` scala
> val classLoader = new URLClassLoader(Array(compileDir.toURI.toURL) ++ compilationClassPaths)
> ```
> 
> and the `compilationClassPaths` constructor parameter should be an `Iterable[URL]`.
> 
> </details>
> Your second change won't work as we already tried it and it needs the parent classpath.  I've addressed your first concern though, let me know what you think.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/108)_

jmorra:
Reviewed 2 of 4 files at r1, 2 of 2 files at r2.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/108)_

gpetrosk:
Tested offline scorer and verified change still loads h2o models correctly

PR #110, opened 2016-09-02
Closes #109

### LINKED ISSUE #109: Adapt Semantics to a new Semantics with the same implementation but different type parameters

deaktator:
Initial `MorphableSemantics` interface.  Needs additional documentation.  Also need to retrofit existing `Semantics` implementations to extends `MorphableSemantics`.

---

This change is [Reviewable](https://reviewable.io/reviews/eharmony/aloha/110)


aloha-core/src/test/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala
```diff
@@ -0,0 +1,98 @@
...
+    * [[MorphableSemantics]].  It can be the case that a specific implementation of
+    * [[MorphableSemantics]] doesn't allow for all `B`, in which case a `None` can be returned.
+    * @param ri reflection information that may be necessary to determine whether to create
+    *           the [[MorphableSemantics]] that was requested.
+    * @tparam B input type for the new [[MorphableSemantics]] instance that might be created.
+    *           A [[MorphableSemantics]] instance may choose not allow morphing to all `B`.
+    *           In that case, a `None` will be returned.
+    * @return
+    */
+  def morph[B](implicit ri: RefInfo[B]): Option[MorphableSemantics[M, B]]
```
deaktator: Should we use context bounds or an implicit `RefInfo[B]` parameter here?

aloha-core/src/main/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala
```diff
@@ -0,0 +1,98 @@
...
+    * [[MorphableSemantics]].  It can be the case that a specific implementation of
+    * [[MorphableSemantics]] doesn't allow for all `B`, in which case a `None` can be returned.
+    * @param ri reflection information that may be necessary to determine whether to create
+    *           the [[MorphableSemantics]] that was requested.
+    * @tparam B input type for the new [[MorphableSemantics]] instance that might be created.
+    *           A [[MorphableSemantics]] instance may choose not allow morphing to all `B`.
+    *           In that case, a `None` will be returned.
+    * @return
+    */
+  def morph[B](implicit ri: RefInfo[B]): Option[MorphableSemantics[M, B]]
```
deaktator, line 67: Should this use context bounds instead of an implicit `RefInfo[B]` parameter?  Who has any thoughts?

jmorra:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala, line 97 [r2]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQh9NZMKatqQr6NbA0v-r2-97:-KQqEYZrm4FKZAoZOnAZ:b-w247z) ([raw file](https://github.com/eharmony/aloha/blob/ba8d3495adeb7c8b78376c185dbd5c2c3c8012f2/aloha-core/src/main/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala#L97)):_
<details><summary>

> Should this use context bounds instead of an implicit `RefInfo[B]` parameter?  Who has any thoughts?
> </details>
> I would think that we should use `RefInfo` as that's what we'll have in the VWJNIModel right?

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala, line 61 [r3]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQqEEjNsa-GRBbd2n_y:-KQqEEjNsa-GRBbd2n_z:bym4w5h) ([raw file](https://github.com/eharmony/aloha/blob/a09853e429b74024c9ce9caf78f113e171d7251d/aloha-core/src/main/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala#L61)):_

> ``` Scala
>     * {{{
>     * case class X[A]() extends Semantics[A] with MorphableSemantics[X, A] {
>     *   def semantics: X[A] = this
> ```

Didn't we play around with something like `this.type`.  I know you ended up not liking it, but do you think it's worth referencing in a comment here?

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala, line 220 [r3]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQqErhsWFJ8ipBO0KKT:-KQqErhsWFJ8ipBO0KKU:b205pdg) ([raw file](https://github.com/eharmony/aloha/blob/a09853e429b74024c9ce9caf78f113e171d7251d/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala#L220)):_

> ``` Scala
>     // way to do this check.
>     // TODO: Find a better way to do the check in second term of the filter.
>     Option(plugin) filter { p =>
> ```

Isn't this exactly what we were trying to avoid in the VWJNIModel?  If we have to do this then we should think about a better way.  We'll still need to add a line for each plugin we'll eventually support.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

deaktator:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala, line 97 [r2]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQh9NZMKatqQr6NbA0v-r2-97:-KQqGOcNj7XEphMv-DOp:b32w0x) ([raw file](https://github.com/eharmony/aloha/blob/ba8d3495adeb7c8b78376c185dbd5c2c3c8012f2/aloha-core/src/main/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala#L97)):_
<details><summary>

> I would think that we should use `RefInfo` as that's what we'll have in the VWJNIModel right?
> </details>
> OK

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala, line 61 [r3]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQqEEjNsa-GRBbd2n_y:-KQqGF-dFGTYXmseyvMQ:b-72rcfm) ([raw file](https://github.com/eharmony/aloha/blob/a09853e429b74024c9ce9caf78f113e171d7251d/aloha-core/src/main/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala#L61)):_
<details><summary>

> Didn't we play around with something like `this.type`.  I know you ended up not liking it, but do you think it's worth referencing in a comment here?
> </details>
> OK.  Will add.

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala, line 220 [r3]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQqErhsWFJ8ipBO0KKT:-KQqGQziVKfONK2puL03:b-nk8tbz) ([raw file](https://github.com/eharmony/aloha/blob/a09853e429b74024c9ce9caf78f113e171d7251d/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala#L220)):_
<details><summary>

> Isn't this exactly what we were trying to avoid in the VWJNIModel?  If we have to do this then we should think about a better way.  We'll still need to add a line for each plugin we'll eventually support.
> </details>
> On it.  I have a way to do it that avoids the match statement

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

deaktator:
Reviewed 1 of 2 files at r2, 4 of 4 files at r4.

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/MorphableCompiledSemanticsPlugin.scala, line 13 [r4]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQt1evj8yKesod-JmrR:-KQt1evj8yKesod-JmrS:b-dobbgs) ([raw file](https://github.com/eharmony/aloha/blob/0696580fd1bf9194c4cbff97e0bf341a01642a01/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/MorphableCompiledSemanticsPlugin.scala#L13)):_

> ``` Scala
>   * @author deaktator
>   */
> trait MorphableCompiledSemanticsPlugin {
> ```

Something about this rubs me the wrong way.  

I kind of feel like the contract in `morph` should obligate the subtypes of `CompiledSemanticsPlugin` to return the same type but parameterized by the type `B`.  This signature for `morph` is adequate but not great.  The downside is it's hard to use but maybe we should just do it and suck it up.  I'll explore this through testing.  

The counter argument is two fold: 
1. Once the `CompiledSemanticsPlugin` is provided to the `CompiledSemantics` in the constructor, it's type is lost.  
2. Maybe a type has other embedded types and it would unnecessarily limit the flexibility of implementations. 

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

deaktator:
Reviewed 1 of 1 files at r5.

---

_[aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala, line 2 [r5]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQtMeskMIZrWevWWLgA:-KQtMeskMIZrWevWWLgB:bfyn4qt) ([raw file](https://github.com/eharmony/aloha/blob/50361865f3f5df186f464e7af7b3d41a6a14106f/aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala#L2)):_

> ``` Scala
> package com.eharmony.aloha.semantics.compiled
> ```

How does this look?  This shows that everything works and it can be adapted for use in other code (e.g., VW JNI model for CDADF).

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

deaktator:
Reviewed 1 of 2 files at r2, 3 of 3 files at r6.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala, line 51 [r6]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQwugCwJp7B2CQ9xh3O:-KQwugCwJp7B2CQ9xh3P:b-h5usuv) ([raw file](https://github.com/eharmony/aloha/blob/0eca85bdfa15ddb6c86ab13ea9069ca9e1bd6412/aloha-core/src/main/scala/com/eharmony/aloha/semantics/MorphableSemantics.scala#L51)):_

> ``` Scala
>   * @author deaktator
>   */
> trait MorphableSemantics[M[_] <: Semantics[_], A] extends Semantics[A] {
> ```

I like this better.

---

_[aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala, line 59 [r6]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQwuneOsD1S9QNLWj2J:-KQwuneOsD1S9QNLWj2K:bingqm2) ([raw file](https://github.com/eharmony/aloha/blob/0eca85bdfa15ddb6c86ab13ea9069ca9e1bd6412/aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala#L59)):_

> ``` Scala
>   private[this] def morphedSemantics[B: RefInfo](s: Semantics[UserProto]): Semantics[B] = {
>     val sO: Option[Semantics[B]] = s match {
>       case ms: MorphableSemantics[_, _] => ms.morph[B]
> ```

Looking at this pattern where we throw away the type parameters makes me think that we should have an additional interface that doesn't require parameterization and can only be mixed into `MorphableSemantics`.  Perhaps something like: 

``` scala
trait EasyMorphableSemantics { self: MorphableSemantics[_, _] => 
  def easyMorph[B: RefInfo]: Option[Semantics[B]] = self.morph[B]
}
```

Then matching is easier: 

``` scala
case s: EasyMorphableSemantics => s.easyMorph[B]
```

What do you think?

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

deaktator:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

_[aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala, line 59 [r6]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQwuneOsD1S9QNLWj2J:-KQwwM-yEt6OfB4iyjfG:b2qludb) ([raw file](https://github.com/eharmony/aloha/blob/0eca85bdfa15ddb6c86ab13ea9069ca9e1bd6412/aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala#L59)):_
<details><summary>

> Looking at this pattern where we throw away the type parameters makes me think that we should have an additional interface that doesn't require parameterization and can only be mixed into `MorphableSemantics`.  Perhaps something like: 
> 
> ``` scala
> trait EasyMorphableSemantics { self: MorphableSemantics[_, _] => 
>   def easyMorph[B: RefInfo]: Option[Semantics[B]] = self.morph[B]
> }
> ```
> 
> Then matching is easier: 
> 
> ``` scala
> case s: EasyMorphableSemantics => s.morph[B]
> ```
> 
> What do you think?
> </details>
> Or maybe it doesn't need any methods at all.  But it could be that both interfaces require the other as a self-type.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

jmorra:

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala, line 220 [r3]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQqErhsWFJ8ipBO0KKT:-KR-AUS7E7oPzfCcKYeL:bofpvnq) ([raw file](https://github.com/eharmony/aloha/blob/a09853e429b74024c9ce9caf78f113e171d7251d/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala#L220)):_
<details><summary>

> On it.  I have a way to do it that avoids the match statement
> </details>
> So if I understand this, we still need the switch for the semantics just not for the plugin right?

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala, line 5 [r6]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KR-ANu0aY8riSg8M6-m:-KR-ANu1GW38uiHd8jXf:b-v1ejkl) ([raw file](https://github.com/eharmony/aloha/blob/0eca85bdfa15ddb6c86ab13ea9069ca9e1bd6412/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala#L5)):_

> ``` Scala
> import com.eharmony.aloha.semantics.compiled.plugin.MorphableCompiledSemanticsPlugin
> import com.eharmony.aloha.semantics.compiled.plugin.proto.CompiledSemanticsProtoPlugin
> import com.google.protobuf.GeneratedMessage
> ```

Now we shouldn't need the proto imports right?

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/MorphableCompiledSemanticsPlugin.scala, line 13 [r4]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQt1evj8yKesod-JmrR:-KR-AkwDjVxRhK4zMP_v:bw590r6) ([raw file](https://github.com/eharmony/aloha/blob/0696580fd1bf9194c4cbff97e0bf341a01642a01/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/MorphableCompiledSemanticsPlugin.scala#L13)):_
<details><summary>

> Something about this rubs me the wrong way.  
> 
> I kind of feel like the contract in `morph` should obligate the subtypes of `CompiledSemanticsPlugin` to return the same type but parameterized by the type `B`.  This signature for `morph` is adequate but not great.  The downside is it's hard to use but maybe we should just do it and suck it up.  I'll explore this through testing.  
> 
> The counter argument is two fold: 
> 1. Once the `CompiledSemanticsPlugin` is provided to the `CompiledSemantics` in the constructor, it's type is lost.  
> 2. Maybe a type has other embedded types and it would unnecessarily limit the flexibility of implementations. 
>    </details>
>    Let me know what you decide to do here

---

_[aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala, line 59 [r6]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQwuneOsD1S9QNLWj2J:-KR-BPLSoas3nlT0IZRA:b-jl9n6i) ([raw file](https://github.com/eharmony/aloha/blob/0eca85bdfa15ddb6c86ab13ea9069ca9e1bd6412/aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala#L59)):_
<details><summary>

> Or maybe it doesn't need any methods at all.  But it could be that both interfaces require the other as a self-type.
> </details>
> I don't really like the easy ideas cause I think it complicates matters for the user.  If there are 2 ways of doing things then the user has to think of which one is better and why.  One way is cleaner IMO.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

deaktator:

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala, line 220 [r3]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQqErhsWFJ8ipBO0KKT:-KR-BtfsfIoCTwhEScNw:b5zre59) ([raw file](https://github.com/eharmony/aloha/blob/a09853e429b74024c9ce9caf78f113e171d7251d/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala#L220)):_
<details><summary>

> So if I understand this, we still need the switch for the semantics just not for the plugin right?
> </details>
> Yup.  Unfortunately.  I don't want to pull the `morph` method into `Semantics` because that would necessitate a change in the class signature which I don't want to do because it would introduce higher-kinded types into the top level construction of `Semantics` and that would be a nightmare on the java side.

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala, line 5 [r6]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KR-ANu0aY8riSg8M6-m:-KR-CdGpC9UXZq4BNaza:b-jwriuk) ([raw file](https://github.com/eharmony/aloha/blob/0eca85bdfa15ddb6c86ab13ea9069ca9e1bd6412/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala#L5)):_
<details><summary>

> Now we shouldn't need the proto imports right?
> </details>
> Nope.  That is an oversight.  I will delete.

---

_[aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala, line 59 [r6]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQwuneOsD1S9QNLWj2J:-KR-DPbAbLEWHPyc3poG:b-y300br) ([raw file](https://github.com/eharmony/aloha/blob/0eca85bdfa15ddb6c86ab13ea9069ca9e1bd6412/aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala#L59)):_
<details><summary>

> I don't really like the easy ideas cause I think it complicates matters for the user.  If there are 2 ways of doing things then the user has to think of which one is better and why.  One way is cleaner IMO.
> </details>
> We could still do it and make the non-easy version `protected[this]` which would allow only one publicly accessible way.  I'm not sure that this worth it, but I think we could make it work.  Let's talk.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

jmorra:
Reviewed 1 of 2 files at r2, 2 of 4 files at r4, 3 of 3 files at r6.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

jmorra:
Reviewed 1 of 1 files at r7.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

deaktator:
Reviewed 1 of 1 files at r7.

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala, line 1 [r7]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KR-KmxCbj3H2-JRsYyw:-KR-KmxCbj3H2-JRsYyx:bjo7ubi) ([raw file](https://github.com/eharmony/aloha/blob/6621fd1e42e46a1d6e1e7d016ede08c6c567fffb/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/CompiledSemantics.scala#L1)):_

> ``` Scala
> package com.eharmony.aloha.semantics.compiled
> ```

Optimized imports.

---

_[aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/MorphableCompiledSemanticsPlugin.scala, line 13 [r4]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQt1evj8yKesod-JmrR:-KR-KsW5FxGjT4hJSZxW:bbrjrux) ([raw file](https://github.com/eharmony/aloha/blob/0696580fd1bf9194c4cbff97e0bf341a01642a01/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/MorphableCompiledSemanticsPlugin.scala#L13)):_
<details><summary>

> Let me know what you decide to do here
> </details>
> I am going to leave as is.  I was thinking that if we had a POJO plugin that did POJOs and a POJO had an embedded protocol buffer, we might want to be able to adapt to a protocol buffer semantics.

---

_[aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala, line 59 [r6]](https://reviewable.io:443/reviews/eharmony/aloha/110#-KQwuneOsD1S9QNLWj2J:-KR-LA0U3KsTguTDMlz7:b-aem8ck) ([raw file](https://github.com/eharmony/aloha/blob/0eca85bdfa15ddb6c86ab13ea9069ca9e1bd6412/aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/MorphableSemanticsTest.scala#L59)):_
<details><summary>

> We could still do it and make the non-easy version `protected[this]` which would allow only one publicly accessible way.  I'm not sure that this worth it, but I think we could make it work.  Let's talk.
> </details>
> I'm going to pass on this.  It's not necessary and the only ones impacted are Aloha _developers_ so I don't mind it being a little trickier but with a cleaner interface.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

deaktator:
Reviewed 1 of 1 files at r8.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

jmorra:
Reviewed 1 of 1 files at r8.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/eharmony/aloha/110)_

PR #113, opened 2016-09-17
closes #112

### LINKED ISSUE #112: Add load testing functionality to aloha-cli
This should allow the user to specify via CLI the ability to load test aloha models.  This should not output predictions but rather information about speed and memory usage.  This should enable by default multithreaded scoring and should output something like:
- _unix time at the time of report_
- _# running predictions_
- _# running empty predictions_
- _memory usage_
- _throughput_

Users should be able to specify input data like in current scoring CLI as well as 
- _number of rounds to loop over input_
- _how often to report_

deaktator:
Added load test functionality to modelrunner CLI.

---

This change is [Reviewable](https://reviewable.io/reviews/eharmony/aloha/113)

### DIFF
--- a/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
+++ b/aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
@@ -1,20 +1,25 @@
 package com.eharmony.aloha.semantics.compiled.plugin.csv
 
-import java.io.{File, PrintStream}
+import java.io.{InputStream, OutputStream, File, PrintStream}
 import java.util.regex.Matcher
 
 import com.eharmony.aloha
 import com.eharmony.aloha.annotate.CLI
 import com.eharmony.aloha.factory.ModelFactory
 import com.eharmony.aloha.io.StringReadable
+import com.eharmony.aloha.models.Model
 import com.eharmony.aloha.reflect.{RefInfo, RefInfoOps}
 import com.eharmony.aloha.score.conversions.ScoreConverter
 import com.eharmony.aloha.score.conversions.ScoreConverter.Implicits._
 import com.eharmony.aloha.semantics.compiled.{CompiledSemanticsPlugin, CompiledSemantics}
 import com.eharmony.aloha.semantics.compiled.compiler.TwitterEvalCompiler
 import com.eharmony.aloha.semantics.compiled.plugin.proto.CompiledSemanticsProtoPlugin
+import com.eharmony.aloha.util.Timing
 import com.google.protobuf.GeneratedMessage
 import org.apache.commons.codec.binary.Base64
+
+import scala.collection.parallel.immutable.ParVector
+
 // import org.apache.commons.lang3.StringEscapeUtils
 import org.apache.commons.vfs2.{FileObject, VFS}
 import spray.json.{JsonFormat, pimpString}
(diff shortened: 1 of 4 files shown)


deaktator:
Currently scoring in the _modelrunner_ CLI uses the model's `apply` function rather than the `score` function.  This is simpler but doesn't keep track of missing features, error messages, etc.  I was wondering if I should add an option to return Aloha score protos instead of the the raw values.  The scores protos would be discarded, but I was thinking this could be useful when performing **load test** functionality.  Being able to toggle between raw and score object predictions in the CLI would allow users to choose the score type more representative, in terms of speed and memory, of the user's production environment.

What do you think @jmorra ?  Is this worthwhile or just a waste of time?

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
```diff
@@ -179,7 +184,126 @@ object OutputType extends Enumeration {
...
+
+            // Create a writing function.  The first function parameter is the input, the second is the model output.
+            val writingFunc: (String, String) => String = inputPosInOutput match {
+                case InputPosition.Neither => (i, o) => o
+                case InputPosition.Before => (i, o) => s"$i$outputSep$o"
+                case InputPosition.After => (i, o) => s"$o$outputSep$i"
+                case InputPosition.Both => throw new IllegalStateException("Can't output input before and after output.")
+            }
+
+            in.foreach{ line => pStream.println(writingFunc(line, predictionFn(line).map(_.toString).getOrElse(predictionMissing))) }
```
jmorra: fold?

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
```diff
@@ -179,7 +184,126 @@ object OutputType extends Enumeration {
...
+        val t = outputSep
+        s"loop_number${t}pred_sec${t}running_pred${t}running_nonempty_pred${t}mem_used_mb${t}mem_unallocated_mb${t}unixtime_ns"
+    }
+
+    private[this] def report(loops: Int, pred: Long, nonEmpty: Long, intervalTime: Float, intervalSize: Int, models: Int): String = {
+        val time = System.nanoTime()
+        val t = outputSep
+        val predSec = intervalSize * models / intervalTime
+        val (usedMB, unallocMB) = memStats()
+        s"$loops$t$predSec$t$pred$t$nonEmpty$t$usedMB$t$unallocMB$t$time"
```
jmorra, line 281: Maybe more legible as `Seq(...).mkString("\t")`

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
```diff
@@ -179,7 +184,126 @@ object OutputType extends Enumeration {
...
+    }
+
+    private[this] def memStats() = {
+        val r = Runtime.getRuntime
+        val tot = r.totalMemory()
+        val free = r.freeMemory()
+        val max = r.maxMemory()
+        val used = tot - free
+        val unalloc = max - used
+        (used / (1 << 20).toFloat, unalloc / (1 << 20).toFloat)
```
jmorra, line 291: Are you reporting in MB (for the bitshift of 20)?  Maybe name the function something more legible if that's the case.

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
```diff
@@ -528,21 +691,35 @@ object CsvModelRunner {
...
-
-        (conf, in, fn, out, closeOut)
+    def getPredictionOutputFormat(config: Option[CsvModelRunnerConfig]): Option[PredictionOutputFormat] = config map { case conf =>
+        val inputType: InputType = conf.inputType.right.get.get
+
+        // (String) => Nothing
+        // Model[Nothing, Any]
+        val (inF, model) = inputAndModel(inputType, conf.outputType, conf.imports, conf.classCacheDir, conf.model.get)
+
+        val out: OutputStream = conf.outputFile.map(f => VFS.getManager.resolveFile(f).getContent.getOutputStream).getOrElse(System.out)
```
jmorra: fold

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
```diff
@@ -528,21 +691,35 @@ object CsvModelRunner {
...
+        val closeOut: Boolean = conf.outputFile.isDefined
+
+        val is: InputStream = conf.inputFile.map(f => VFS.getManager.resolveFile(f).getContent.getInputStream).getOrElse(System.in)
```
jmorra: fold

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
```diff
@@ -528,21 +691,35 @@ object CsvModelRunner {
...
+        val in: Iterator[String] = scala.io.Source.fromInputStream(is).getLines()
```
jmorra, line 716: The source is never closed.

jmorra (changes requested):
Made a few requests for changes, but overall I think this looks good.

jmorra:
I think your idea about doing score based load testing as well is a good idea.  Because that is frequently used in production it'll be more representative of what is expected.  Do you think it's also worthwhile to add an option for model closing/opening.  We've had issues with memory leaks in models before and I could see a real useful flag being something like "close and reopen the model every N iterations".  Maybe that could also be it's own testing functionality.  What do you think?

pom.xml
```diff
@@ -300,6 +300,7 @@
                             <include>**/*Test.scala</include>
                         </includes>
                         <forkMode>once</forkMode>
+                        <argLine>-Xmx2g</argLine>
```
jmorra, line 303: This was needed to get the build to pass again on Travis.

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
deaktator: Sure.  Will Do.

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
deaktator: OK.  Fine by me.

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
deaktator: Will change name.

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
deaktator: You and your folds (Odersky hates folds).  Will Change.

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
deaktator: Again?  Geez.  OK.  Will change

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
deaktator: Yeah it is sucker!  https://github.com/scala/scala/blob/v2.10.5/src/library/scala/io/Source.scala#L96-Lundefined

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/csv/CsvModelRunner.scala
deaktator: I checked this out and it won't work with The `Seq` /  `mkString` version.  That's because the values are downcast to AnyVal and consequently, the formatting of the integral values appears as if they are doubles.  Reverting back.

jmorra (approved):
These look good to me.

deaktator (commented):
Looked.  OK.

deaktator (commented):
Done.

PR #123, opened 2016-12-19
Closes #8

### LINKED ISSUE #8: Make auditor output type
Remove the notion of a ScoreProto and replace that with a recursive data type that allows representation of all the subscores.

deaktator:
Changing API to use auditors.  Model trait will be vastly simplified.  Something like:

```scala
trait Model[-A, +B] extends (A => B)  // Main model API: just a function
trait TypeCtor { type TC[+A] }
trait AuditedModel[T <: TypeCtor, N, -A, +B <: T#TC[N]] extends Model[A, B] {
  private[aloha] def typeConstructed: AuditedModel[T, N, A, T#TC[N]] =
    this.asInstanceOf[AuditedModel[T, N, A, T#TC[N]]]
}
```

Still working out the APIs, variance, etc.  The nice thing is that everything works nicely in Java:

```java
import com.eharmony.aloha.score.audit.OptionAuditor.OptionTC;

@SuppressWarnings("unchecked")
static <A> Manifest<A> manifest(final String strRep) {
  return (Manifest<A>) ManifestParser.parse(strRep).right().get();
}


// Easy to create Auditors from Java!
Manifest<Integer> refInfoInt = manifest(Integer.class.getCanonicalName());
OptionAuditor<Integer> intAud = new OptionAuditor<>(refInfoInt);

Manifest<Float> refInfoFloat = manifest(Float.class.getCanonicalName());
OptionAuditor<Float> floatAud = new OptionAuditor<>(refInfoFloat);


// Easy to directly instantiate models.  Last two type parameters are the Model 
// type parameters.
ConstantModel<OptionTC, Integer, Object, Option<Integer>> cModel =
    new ConstantModel<>(ModelId.empty(), 1, intAud);

// Last two type parameters are the Model type parameters.
HierarchicalConstantModel<OptionTC, Integer, Float, Object, Option<Float>> hcm =
    HierarchicalConstantModel.createFromJava(ModelId.empty(), 2f, floatAud, cModel);

// Seamless downcasting to the model interface.
Model<Object, Option<Float>> m = hcm;

// Easy to call prediction function:
Option<Float> result = m.apply(null);    // null OK b/c model just returns a constant.
```

and of course, Scala works nicely:

```scala
val cm = ConstantModel(ModelId(), 1, OptionAuditor[Int])
val hcm = HierarchicalConstantModel(ModelId(), 2f, OptionAuditor[Float])(cm)
val hcmModel: Model[Any, Option[Float]] = hcm
assert( Option(2f) == hcm( () ) )
```

---
This change is [Reviewable](https://reviewable.io/reviews/eharmony/aloha/123)


deaktator:
Currently, [take3](https://github.com/eHarmony/aloha/tree/8-auditor/aloha-core/src/test/scala/com/eharmony/aloha/score/audit/take3) is the API I like the best.  I'd like to make this the de facto one.  It has lots of type parameters (some are redundant) but it works the best in Java and Scala.  It will work completely with type inference, even in Java.  For instance, even the `TypeCtor` interface is entirely for Java compatibility.  It could just be a higher kinded-type, but this doesn't work in Java.

aloha-core/src/main/scala/com/eharmony/aloha/score/audit/Auditor.scala
```diff
@@ -0,0 +1,42 @@
+package com.eharmony.aloha.score.audit
+
+import com.eharmony.aloha.reflect.RefInfo
+
+import scala.language.higherKinds
+
+/**
+  *
+  * Created by deaktator on 12/9/16.
```
jmorra: Can you add a description here about what an Auditor is.

aloha-core/src/main/scala/com/eharmony/aloha/score/audit/EitherAuditor.scala
```diff
@@ -0,0 +1,32 @@
...
+import com.eharmony.aloha.score.audit.EitherAuditor.{Failure, Result, Success}
+
+/**
+  * Created by ryan on 12/12/16.
+  */
+final case class EitherAuditor[K, V](implicit val refInfo: RefInfo[V])
+  extends MorphableAuditor[K, V, Result[K, V], EitherAuditor[K, V]] {
+  override type AuditOutput[A] = Result[K, A]
+  override def auditor[U: RefInfo]: Option[EitherAuditor[K, U]] = Option(EitherAuditor[K, U]())
+//  override private[aloha] def failure[C](key: K, errorMsgs: => Seq[String], missingVarNames: => Set[String], childValues: Seq[Result]): Failure[K] = Failure(key, errorMsgs, missingVarNames, childValues)
```
jmorra: Remove dead code.

jmorra (changes requested):
This looks awesome.  I love the docs and the test.  But I can't help but to think I'm missing something.  Shouldn't there now be a new Model trait?  And the old Model trait should be removed?  Also this should become version 4.  It's a HUGE change.

deaktator:
@jmorra The only thing to look at right now is the `take3` package in `src/test`.  This is self-contained and describes what the API will be.  The rest of the code is not yet done.  I just had to work through a bunch of different ways to accomplish the same thing and figure out what worked the best.  The `take3` one is very nice and doesn't required the specification of any types as evidenced by the first comment in the PR.  

The model factories will return a `Model[A, B]` which is really elegant.  It's just a Scala function and it will also have a `ModelIdentity`.

This took a long time to get right on the Java side.  I'll probably write up a blog article to describe the necessity of all of the components (especially the `TypeCtor1`).

Definitely agree on the version bump.  This is a definitely a major revision.

aloha-core/src/test/java/com/eharmony/aloha/score/audit/take3/JavaTestTake3.java
```diff
@@ -0,0 +1,132 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+/**
+ * Shows how
+ * Created by ryan on 12/16/16.
+ */
+@RunWith(BlockJUnit4ClassRunner.class)
+public class JavaTestTake3 {
```
jmorra: This looks very nice.  Could you show an example instantiating from a JSON through the factory?

deaktator: I will.  That's not done yet.  I'm thinking something like: 

```java
OptionAuditor<Integer> aud = new OptionAuditor<>(refInfoInt);
Semantics<MyDataType> sem = ...;
ModelFactory<OptionTC, Integer, MyDataType, Option<Integer>> factory = 
    ModelFactory.createFromJava(..., sem, aud, ...);
File jsonFile = ...;

// Get model from factory.
Try<Model<MyDataType, Option<Integer>>> modelTry = factory.fromFile(jsonFile);
```

deaktator:
Just made a huge commit.  Everything is compiling at least.  Had to change every place where models and factories are defined, instantiated, and used.  Still trying to make the unit tests pass.

deaktator:
There still needs to be a fair amount of cleanup including looking at the documentation and modifying it.  I'm not going to try to rip out the protobuf stuff to a separate module in this PR.  That'll have to be a separate issue.

aloha-core/src/main/scala/com/eharmony/aloha/audit/Auditor.scala
```diff
@@ -0,0 +1,21 @@
+package com.eharmony.aloha.audit
+
+import com.eharmony.aloha.id.ModelIdentity
+
+/**
+  * Created by ryan on 1/11/17.
+  */
+trait Auditor[U, -N, +B <: U] {
```
vsuthichai: Would it be possible to lengthen how the type parameters are named?  Scala does emphasize brevity, but I find it a little difficult to decipher what the type parameters are from first glance.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/scoreproto/ScoreAuditor.scala
```diff
@@ -0,0 +1,254 @@
...
+      val m = mId(key)
+      val b = Score.newBuilder
+      if (subValues.nonEmpty)
+        b.addAllSubScores(asJavaIterable(subValues))
+      val mm = accumulateMissing(missingVarNames, subValues)
+      val e = errorMsgs
+      if (mm.nonEmpty || e.nonEmpty) {
+        val eb = ScoreError.newBuilder.
+          setMissingFeatures(MissingRequiredFields.newBuilder.addAllNames(asJavaIterable(mm))).
+          setModel(m)
```
vsuthichai: Why is the model id and name only set in the error flow?  Does this pertain to the submodel identity during a recursive tree traversal?

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/scoreproto/ScoreAuditor.scala
```diff
@@ -0,0 +1,254 @@
...
+                               missingVarNames: => Set[String],
+                               subValues: Seq[Score]): Score = {
+      val b = Score.newBuilder
+      if (subValues.nonEmpty)
+        b.addAllSubScores(asJavaIterable(subValues))
+      val e = ScoreError.newBuilder.setModel(mId(key)).addAllMessages(asJavaIterable(errorMsgs))
+      val m = accumulateMissing(missingVarNames, subValues)
+      if (m.nonEmpty)
+        e.setMissingFeatures(MissingRequiredFields.newBuilder.addAllNames(asJavaIterable(m)))
+      b.setError(e).build
```
vsuthichai: Is this `Score` proto eventually somehow propagated to the next auditor and becomes a part of the `subValues` argument?  What happens to the returned `Score` here?

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/scoreproto/ScoreAuditor.scala
```diff
@@ -0,0 +1,254 @@
...
+/**
+  * Created by ryan on 1/17/17.
+  */
+// TODO: Move to separate module.
+sealed abstract class ScoreAuditor[N] extends MorphableAuditor[Score, N, Score]
+
+object ScoreAuditor {
+
+  def apply[N](implicit ri: RefInfo[N]): Option[ScoreAuditor[N]] = {
+    ri match {
```
vsuthichai: Like the factory method.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/scoreproto/ScoreAuditor.scala
```diff
@@ -0,0 +1,254 @@
...
+
+  private[this] object ByteAuditor extends ScoreAuditorImpl[Byte] {
+    override def boxScore(valueToAudit: Byte): BaseScore.Builder =
+      BaseScore.newBuilder.setType(ScoreType.INT).setExtension(
+        IntScore.impl,
+        IntScore.newBuilder.setScore(valueToAudit).build
+      )
+
+    private[aloha] def unapply(value: Score): Option[Byte] = {
+      val v = value.getScore.getExtension(IntScore.impl)
```
vsuthichai: protobufs have no support for bytes or shorts?

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/scoreproto/ScoreAuditor.scala
```diff
@@ -0,0 +1,254 @@
...
+import com.eharmony.aloha.score.Scores.Score.BaseScore.ScoreType
+import com.eharmony.aloha.score.Scores.Score._
+
+import scala.collection.JavaConversions.{asJavaIterable, iterableAsScalaIterable}
+
+/**
+  * Created by ryan on 1/17/17.
+  */
+// TODO: Move to separate module.
+sealed abstract class ScoreAuditor[N] extends MorphableAuditor[Score, N, Score]
```
vsuthichai: I assume once we move away from protobufs, the Score protos will be replaced by some other type, and there would be some other ScoreAuditor type for non-protos.

Would it make sense to rename this to ScoreProtoAuditor instead?

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/scoreproto/ScoreAuditor.scala
vsuthichai: Looked at the docs, apparently not.

aloha-core/src/main/scala/com/eharmony/aloha/algebra/NaturalOptionTransformation.scala
```diff
@@ -0,0 +1,10 @@
+package com.eharmony.aloha.algebra
+
+import scala.language.higherKinds
+
+/**
+  * Created by ryan on 1/25/17.
+  */
+trait NaturalOptionTransformation[F[_], G[_]] {
```
vsuthichai, line 8: If I understand this correctly, this is to convert one higher kinded type to an option of another higher kinded type.  Why is it needed?

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/scoreproto/ScoreAuditor.scala
```diff
@@ -0,0 +1,254 @@
...
+        IntScore.newBuilder.setScore(valueToAudit).build
+      )
+
+    private[aloha] def unapply(value: Score): Option[Byte] = {
+      val v = value.getScore.getExtension(IntScore.impl)
+      if (v.hasScore) {
+        val sc = v.getScore
+        if (Byte.MinValue <= sc && sc <= Byte.MaxValue)
+          Some(sc.toByte)
+        else None
```
vsuthichai: Do we want to log this or propogate this somehow?  It seems like a silent error.  If the model has a score outside the intended values, would this indicate a deeper problem..

vsuthichai:
The last commit has been **far** more legible to me, though it's still taking some effort to humanly ingest all the type parameters.  From what I understand while reading the code, the auditors' whole purpose is to recursively aggregate the scores and errors together to return to the caller.  

Does it make sense to lock down the Auditor trait's `success` and `failure` method signatures with those specific parameters:  `valueToAudit`, `missingVarNames`, etc.  It seems inflexible.  Would passing in some sort of context object that contains that information be better?  Room for growth if needed.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/OptionAuditor.scala
```diff
@@ -0,0 +1,25 @@
+package com.eharmony.aloha.audit.impl
+
+import com.eharmony.aloha.audit.MorphableAuditor
+import com.eharmony.aloha.id.ModelIdentity
+import com.eharmony.aloha.reflect.RefInfo
+
+/**
+  * Created by ryan on 1/30/17.
+  */
+case class OptionAuditor[A]() extends MorphableAuditor[Option[_], A, Option[A]] {
```
vsuthichai, line 10: Not quite sure I understand what this is for.

deaktator:
***Everything is now compiling and all unit tests in all submodules are passing!***

Questions for @vsuthichai @jmorra:

1. Should we delete ensemble models package?  It's never been used and there has never been an ability to create via a model factory.
1. When should we move protos out of `aloha-core`? I want to do this ASAP, but this will affect things like ModelRunnerTool and dataset generation based on protocol buffers.  As a consequence, this will involve a substantial amount of refactoring command line tools.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/OptionAuditor.scala
deaktator: I want to understand what you don't understand.  Can you elaborate?

deaktator:
@vsuthichai How would providing a context to the `Auditor` `success` and `failure` method allow it to be more expandable?  I like the idea but I'm not seeing how this would allow us to inject additional information without rewriting all of the model implementations.

aloha-core/src/main/scala/com/eharmony/aloha/audit/Auditor.scala
```diff
@@ -0,0 +1,28 @@
...
+  *
+  * Created by ryan on 1/11/17.
+  *
+  * @tparam U The upper type bound for items being audited.
+  * @tparam N The "''natural''" output type of a given model.
+  * @tparam B The output type of the Auditor.  This is the same as the output type of a
+  *           [[com.eharmony.aloha.models.Model]].
+  */
+trait Auditor[U, -N, +B <: U] {
+  def failure(modelId: ModelIdentity,
```
jon-morra-zefr, line 16: Why does an auditor need to know about the `ModelIdentity`?

aloha-core/src/main/scala/com/eharmony/aloha/factory/ModelFactory.scala
```diff
@@ -1,225 +1,384 @@
...
+  * @param auditor
+  * @param parsers
+  * @param refInfoToJsonFormat
+  *
+  * @tparam U
+  * @tparam N
+  * @tparam A
+  * @tparam B
+  *
+//  * @param factory an untyped factory the produces models of kind M
```
jon-morra-zefr: grammar

aloha-core/src/main/scala/com/eharmony/aloha/factory/ModelFactory.scala
```diff
@@ -1,225 +1,384 @@
...
+  *         // because the higher kind M[A, B] erases to Object.
+  *         return modelFactory.fromClasspath(path).get();
+  *
+  *         // This would work (but the cast is annoying) and the coder has to know about the required casting.
+  *         // return ((Try< Model< TestProto, Double > >) modelFactory.fromClasspath(path)).get();
+  *     }
+  * }
+  * }}}
+  *
+  * @param semantics
```
jon-morra-zefr: These docs seem missing.

aloha-core/src/main/scala/com/eharmony/aloha/factory/ModelFactory.scala
```diff
@@ -1,225 +1,384 @@
...
+
+  // TODO: Create a custom printer that truncates string to be shorter for reporting
+  private[this] def jsonPrinter: JsValue => String = CompactPrinter
+
+
+  // json: JsValue, fileStack: List[String], auditor: MorphableAuditor[U, N, B]
+  // TODO: Consider using scala.util.control.TailCalls for trampolining to avoid stack overflows.
+  /**
+    *
+    * From old documentation in ModelFactory
```
jon-morra-zefr, line 175: Why do we care about old docs?

aloha-core/src/main/scala/com/eharmony/aloha/factory/ModelFactory.scala
```diff
@@ -1,225 +1,384 @@
...
+    * @param jsonMaybe
+    * @param inlineReader
+    * @param fileStack
+    * @tparam M
+    * @tparam C
+    * @tparam Y
+    * @return
+    */
+  private[this] def parse[M, C <: U, Y](
+      jsonMaybe: Try[JsObject], // TODO: Should be a Try[JsObject]
```
jon-morra-zefr, line 208: It is

jon-morra-zefr (approved):
I've left some comments, but obviously this is a lot.  Let me know if there is something specific that you want me to look at, otherwise this looks good.

jmorra:
To your comment about removing code.  I think that's a good idea.  The direction of the project does not necessitate these models anymore.

I think we've moving the protos out in the next PR, right?

jmorra (approved):
Too much to review in detail, but I've looked at some of it and it looks good.

vsuthichai (approved):
Approved!

vsuthichai (approved):
Approved, take two!

sahil-goyal (approved):

PR #132, opened 2017-03-03
Closes #131

### LINKED ISSUE #131: Create easy to use Avro model factory
To support this, change the `ModelFactory[U,N,A,B]` to `ModelFactoryImpl[U,N,A,B]` and then create a `ModelFactory[A, B]` trait that `ModelFactoryImpl[U,N,A,B]` extends.

deaktator:
This makes does two things.  First, it changes the current `ModelFactory` to `ModelFactoryImpl` and then creates an abstract base class `ModelFactory` that only has two type parameters.  This is very nice as it simplifies the type signature.

The second thing that it does is to allow the creation of Avro factories very easily.  This can be done without a compiler generated `RefInfo` and instead uses a String representation of a `RefInfo` that is parsed.  I know this is absurd, but the factory method `StdAvroModelFactory.apply` that simplifies the creation of Avro factories returns a `scala.util.Try`  This let's the user do the error handling.

### DIFF
--- a/aloha-core/src/main/scala/com/eharmony/aloha/factory/ModelFactory.scala
+++ b/aloha-core/src/main/scala/com/eharmony/aloha/factory/ModelFactory.scala
@@ -19,7 +19,17 @@ import spray.json.{CompactPrinter, JsObject, JsValue, JsonFormat, JsonReader, Ro
 
 import scala.util.{Failure, Success, Try}
 
-
+/**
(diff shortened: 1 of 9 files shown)


jmorra (approved):

deaktator:
@jmorra, should `ModelFactory` be sealed?  I was thinking maybe it should.

jmorra:
Probably

PR #140, opened 2017-03-07
Closes #139

### LINKED ISSUE #139: Additional Avro factory creation method that directly takes a Schema

deaktator:


jon-morra-zefr (approved):
Looks good and you made me feel bad on the phone.

deaktator:
Now that I'm thinking about it.  I don't like how this all works.  I think that the way to go is the [magnet pattern](http://spray.io/blog/2012-12-13-the-magnet-pattern/) which is totally made for such cases like this.  I'm going to proceed with that.

aloha-io-avro/src/main/scala/com/eharmony/aloha/factory/avro/StdAvroModelFactory.scala
```diff
@@ -1,27 +1,35 @@
...
+    * @param conf a factory configuration.  This can also be used via the Magnet
+    *             pattern.  See implicits in the [[FactoryConfig]] companion object
+    *             to see possible parameter combinations.  Also see the Spray
+    *             [[http://spray.io/blog/2012-12-13-the-magnet-pattern/ Magnet pattern]]
+    *             article for more information.  To enable Magnet pattern syntax, see
+    *             [[FactoryConfig.MagnetSyntax]]
+    *
+    * @return a Try of a ModelFactory.
+    */
+  def fromConfig(conf: FactoryConfig): Try[ModelFactory[GenericRecord, Score]] = conf()
```
deaktator: I like this.  I wish this could be the apply method but it can't because of the previously defined one.  This defines the possibility of the Magnet pattern, which I like.

aloha-io-avro/src/main/scala/com/eharmony/aloha/factory/avro/StdAvroModelFactory.scala
```diff
@@ -43,68 +51,20 @@ object StdAvroModelFactory {
     * @return A Try of a ModelFactory that creates models taking `GenericRecord` instances as
     *         input and returns `com.eharmony.aloha.audit.impl.avro.Score` as output.
     */
+  @deprecated(message = "Prefer ", since = "4.0.1")
```
deaktator: I deprecated this because I would like everyone to use the config objects instead.  Then, someday once this is removed, that can be deprecated and moved to `apply`.

aloha-io-avro/src/main/scala/com/eharmony/aloha/factory/avro/factory_config.scala
```diff
@@ -0,0 +1,179 @@
...
+import scala.util.{Failure, Success, Try}
+
+sealed abstract class FactoryConfig {
+  /**
+    * Attempt to create a factory.
+    * @return
+    */
+  def apply(): Try[ModelFactory[GenericRecord, Score]]
+
+  protected[this] def semantics(s: Schema,
```
deaktator, line 31: Placed all of the helper functions into the base class

aloha-io-avro/src/main/scala/com/eharmony/aloha/factory/avro/factory_config.scala
```diff
@@ -0,0 +1,179 @@
...
+import com.eharmony.aloha.semantics.compiled.compiler.TwitterEvalCompiler
+import com.eharmony.aloha.semantics.compiled.plugin.avro.CompiledSemanticsAvroPlugin
+import org.apache.avro.Schema
+import org.apache.avro.generic.GenericRecord
+import org.apache.commons.io.IOUtils
+import scala.concurrent.ExecutionContext.Implicits.global
+
+import scala.util.{Failure, Success, Try}
+
+sealed abstract class FactoryConfig {
```
deaktator: Made an `abstract class` instead of a `trait` so the companion object exposes the proper static functions but that might not be necessary. @jmorra, what do you think?  The only real reason to have the companion object is to support the magnet syntax, via the implicit definitions.

aloha-io-avro/src/main/scala/com/eharmony/aloha/factory/avro/factory_config.scala
```diff
@@ -0,0 +1,179 @@
...
+}
+
+object FactoryConfig {
+
+  /**
+    * To enable the use of the magnet pattern, import `MagnetSyntax._`.
+    * See Spray's [[http://spray.io/blog/2012-12-13-the-magnet-pattern/ Magnet pattern]]
+    * article for more information.
+    */
+  object MagnetSyntax {
```
deaktator: Do we really want to support this?  Seems nice but might provide too many ways to do the same thing.

aloha-io-avro/src/test/scala/com/eharmony/aloha/factory/avro/StdAvroModelFactoryTest.scala
```diff
@@ -17,27 +17,73 @@ import org.junit.runners.BlockJUnit4ClassRunner
...
-    val model: Model[GenericRecord, Score] = factory.fromString(modelJson).get
+    val model: Model[GenericRecord, Score] = factory.fromString(ModelJson).get
+
+    assertEquals(7d, model(record).getValue.asInstanceOf[Double], 0)
+  }
+
+  @Test def testHappyPathUrlMagnetSyntax(): Unit = {
+    // Must import syntax explicitly to use Magnet syntax. Can do
+    // import FactoryConfig.MagnetSyntax._
+    import FactoryConfig.MagnetSyntax.fromUrlCodomainImports
```
deaktator: Should the Magnet syntax be automatic (defined in the implicit scope of FactoryConfig) or should we require explicit imports to enable the syntax?  Or should the syntax be removed entirely as it's not necessary?  The more I think about it, the more I think it should be removed and added later if we choose.  A lot easier to add than remove.

deaktator (commented):
Having second thoughts about the Magnet pattern syntax, but not the configuration objects.  @jmorra what do you think?

aloha-io-avro/src/main/scala/com/eharmony/aloha/factory/avro/factory_config.scala
jon-morra-zefr: I think that's fine.

aloha-io-avro/src/test/scala/com/eharmony/aloha/factory/avro/StdAvroModelFactoryTest.scala
jon-morra-zefr: I agree.

jon-morra-zefr (approved):
Looks good

jmorra (approved):
Looks good

deaktator:
I'm trying to figure out how I made you feel bad on the phone.  I'll have to call and find out (hopefully without doing more damage).

deaktator:
Just had another thought.  This `String`/`Boolean` crap for URLs is a poor abstraction, especially since we already have a better one.  I'm switch it to `aloha.io` package for URLs.

aloha-io-avro/src/main/scala/com/eharmony/aloha/factory/avro/factory_config.scala
```diff
@@ -0,0 +1,165 @@
...
+  * @param imports imports to be injected into feature functions synthesized by the factory.
+  * @param classCacheDir a cache directory on the local machine used to cache class files of
+  *                      the created feature functions used in the models produced by the
+  *                      factory.
+  * @param dereferenceAsOptional whether to treat the dereferencing of repeated variables as
+  *                              an optional type.  This avoids index out of bounds exceptions
+  *                              and is safer but slightly slower.
+  */
+case class UrlConfig(
+    vfs: Vfs,
```
deaktator, line 131: @jmorra This is a big change but I think it's for the best.  It makes things a bit more verbose when defining but I think it's worth it.  I could make the `Vfs1` and `Vfs2` factory methods in `c.e.a.io.vfs` implicit to save a little. 

```scala
object Vfs1 {
  implicit def apply(fo: vfs1.FileObject): Vfs1 = ???
}
```

```scala
object Vfs2 {
  implicit def apply(fo: vfs2.FileObject): Vfs2 = ???
}
```

What do you think?

aloha-io-avro/src/main/scala/com/eharmony/aloha/factory/avro/StdAvroModelFactory.scala
```diff
@@ -43,68 +48,32 @@ object StdAvroModelFactory {
...
-
-  private[this] def refInfo(refInfoStr: String) =
-    RefInfo.fromString(refInfoStr) match {
-      case Left(err) => Failure(new AlohaFactoryException(err))
-      case Right(success) => Success(success.asInstanceOf[RefInfo[Any]])
-    }
-
-  private[this] def schemaInputStream(vfsUrl: String, useVfs2: Boolean) =
-    Try {
+  private[this] def url(modelDomainSchemaVfsUrl: String, useVfs2: Boolean) = {
```
jmorra, line 72: Should this default to `true`?

aloha-io-avro/src/main/scala/com/eharmony/aloha/factory/avro/factory_config.scala
jmorra: I think the way it is is fine.  I like this.

jmorra (approved):
Looks good

PR #152, opened 2017-04-13
Closes #150

### LINKED ISSUE #150: In Avro plugin, variable extraction should be flatMap except for the last component in the path
### Feature Specifcation
```json
{ "name": "ids", "spec": "${x.y.ids} map ind" }
```

### Avro Data
```
record Y   { union { null, array<string> } ids = null; }
record X   { union { null, Y             } y   = null; }
record Msg { union { null, X             } x   = null; }
```

## Desire 
Produce an `Iterable[(String, Double)]` given the feature specification and Avro definitions.

## Problem
The following improper (*pretty-printed*) code is generated:

### Improperly generated code

```scala
{
import com.eharmony.aloha.feature.BasicFunctions._
import scala.math._
identity[GenAggFunc[Msg, Iterable[(String, Double)]]](
  GenFunc.f1(
    GeneratedAccessor(
      "x.y.ids",
      (_0: Msg) => {
        import scala.collection.JavaConversions.asScalaBuffer
        Option(_0.get("x")).asInstanceOf[Option[GenericRecord]].flatMap(_1 =>
          Option(_1.get("y")).asInstanceOf[Option[GenericRecord]].map(_2 =>   // <== !!!
            Option(_2.get("ids")).asInstanceOf[Option[java.util.List[CharSequence]]].map(
              _.toSeq.map(_.toString)))).getOrElse(Nil)
      },
      None
    )
  )(
    "${x.y.ids} map ind",
    (_x__y__ids) => {
     _x__y__ids map ind
    }
  )
)
}
```

The problem lies in the generated code for the `GeneratedAccessor`.  What this really should like is something like the following:

### Expected results (in *for comprehension* style)
```scala
import org.apache.avro.generic.GenericRecord
import com.zefr.avro.training.TrainingRelevancy

val f = (_0: Msg) => {
  import scala.collection.JavaConversions.asScalaBuffer
  (
  for {
    _1 <- Option(_0.get("x")).asInstanceOf[Option[GenericRecord]]
    _2 <- Option(_1.get("y")).asInstanceOf[Option[GenericRecord]]
    _3 <- Option(_2.get("ids")).asInstanceOf[Option[java.util.List[CharSequence]]]
  } yield _3.toSeq.map(_.toString)
  ) getOrElse Nil
}
```

So the problem lies in the improperly generated code emitting a `map` when a `flatMap` should be emitted.

```scala
val g = (_0: Msg) => {
  import scala.collection.JavaConversions.asScalaBuffer
  Option(_0.get("x"))
    .asInstanceOf[Option[GenericRecord]]
    .flatMap(_1 =>
      Option(_1.get("y"))
        .asInstanceOf[Option[GenericRecord]]
        .flatMap(_2 =>          // <== Problem line: Needs to be flatMap
          Option(_2.get("ids"))
            .asInstanceOf[Option[java.util.List[CharSequence]]]
            .map(_.toSeq.map(_.toString))
        )
    ).getOrElse(Nil)
}
```

This is going to be the case for all instances, regardless how long the path to get a variable is.  

**TODO**: Figure out if there are any nuances here or just alway emit `flatMap`.

deaktator:


aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/schemabased/SchemaBasedSemanticsPlugin.scala
```diff
@@ -38,8 +38,12 @@ trait SchemaBasedSemanticsPlugin[A] { self: CompiledSemanticsPlugin[A] =>
 
   private[schemabased] def generateFunction(p: FieldAccessorPartition) = {
     val (before, repeated, after) = p
-    val b = determineMappingPartition(before)
-    val a = determineMappingPartition(after)
+
+    // We need to be mindful of whether the repeated field is nullable.  This affects the
+    // choice of map / flatMap in the last chain prior to the repeated field.
+    val nullableRepeated = repeated.fold(false)(_.field.nullable)
```
jmorra: This could probably do done with `exists(_.field.nullable)`.

jmorra (approved):
This looks great.  I assume this won't work if we have an array<SOME_COMPLEX_TYPE> right?

aloha-core/src/main/scala/com/eharmony/aloha/semantics/compiled/plugin/schemabased/SchemaBasedSemanticsPlugin.scala
deaktator: agreed.  Will change.

deaktator:
@jmorra Correct, this won't work if we have an array<SOME_COMPLEX_TYPE>?  Does work for 

* `array<RECORD_TYPE>`
* `array<int>`
* `array<float>`
* ...

PR #154, opened 2017-04-19
Closes #153

### LINKED ISSUE #153: Optional repeated fields in the middle of a variable path don't work properly
This is related to #150 but different.

## Important!

**Try to fix this inside `SchemaBasedSemanticsPlugin` and not inside the code generators!**  This is important so that it'll work for other schema based semantics plugin implementations.

## Example Avro Schema
```
record Z { int i; }
record Y { Z rz;   union { null, Z } oz = null; }
record X { union { null, array<Y> } ys = null; }
record Msg { union { null, X } x = null; }
```

## Example Aloha Feature Specifications

```json
"(0 /: ${x.ys.oz.i})(_ + _.getOrElse(0))"
```

```json
"${x.ys.rz.i}.sum"
```

These kinds of things should work but don't.  The reason is that because `ys` is nullable, the map that's currently emitted maps over the Option of the list rather than the list itself.  To remedy, this, another map needs to be emitted inside the first map so that elements of the list are mapped over rather than the option containing the list.

deaktator:
Updated code and tests to fix issues with repeated fields in the middle of a variable path.

This also fixes dereferencing of optional lists.


deaktator (commented):
This looks good! 😃

PR #158, opened 2017-05-10
Closes #157

### LINKED ISSUE #157: Change Logging trait to make _logger transient
Change `Logging` trait to make `_logger` `@transient`.  This will limit the possibility of null pointer exceptions when deserializing models that extends `Logging`.

deaktator:
Fixes the `@transient` issue and removes unnecessary unused code.


jmorra (approved):
Meh

PR #167, opened 2017-08-18
Closes #164

### LINKED ISSUE #164: Update to VW JNI to version 8.4.1

deaktator:
Updated VW version to 8.4.1.  

Only added support for 

- `VWScalarLearner`
- `VWProbLearner`
- `VWMulticlassLearner`

These are the types that extends `VWIntLearner` and `VWFloatLearner`.


aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/VwJniModel.scala
```diff
@@ -143,6 +143,44 @@ object VwJniModel
     with VwJniModelJson
     with Logging {
 
+  private[jni] def toLabelFn[N: RefInfo](labels: Option[SimpleTypeSeq]): Int => N = {
+    labels map { ls =>
+      val tc = convFn(ls.refInfo, RefInfo[N])
+      (i: Int) => tc(ls.values(i - 1))
```
jon-morra-zefr: Why are you subtracting 1 here?  Can you add a comment.

jon-morra-zefr (changes requested):
Very small change, but it looks good.

jmorra:
This looks great, but since we're updating dependent ML libraries, can we also move the H2O version to 3.10.3.2.  This is the version we're currently using.

jmorra (approved):
Looks great

PR #175, opened 2017-09-01
Closes #163

### LINKED ISSUE #163: Support multilabel problems

deaktator:
## NOTE

This is a slightly weird PR since this ticket is to be done in stages.  The goal is to review that stages in pieces, then I'll delete the PR and submit another for the next stage.

## Goals

This PR review should address the `MultilabelModel` **case class** and the supporting functions in the companion object (*but not `Parser` in the `MultilabelModel` companion object*).  It should also cover the `multilabel` **package object** and the `MultilabelModelTest`.

## Details

Currently, the `MultilabelModel` will not support label-dependent features.  This infrastructure should be sufficient to wrap most if not all multi-label predictors.

### DIFF
--- a/aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
+++ b/aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
@@ -30,7 +30,7 @@ import scala.collection.{immutable => sci}
   * @tparam B output type of the model.
   */
 // TODO: When adding label-dep features, a Seq[GenAggFunc[K, Sparse]] will be needed.
-// TODO: To create a Seq[GenAggFunc[K, Sparse]], a Semantics[K] will need to derived a from Semantics[A].
+// TODO: To create a Seq[GenAggFunc[K, Sparse]], a Semantics[K] needs to be derived from a Semantics[A].
 // TODO: MorphableSemantics provides this.  If K is *embedded inside* A, it should be possible in some cases.
 case class MultilabelModel[U, K, -A, +B <: U](
     modelId: ModelIdentity,
--- a/aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
+++ b/aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
@@ -45,7 +45,9 @@ import scala.util.{Failure, Success, Try}
   * @param predictorProducer the function produced when calling this function is responsible for
   *                          getting the data into the correct type and using it within an
   *                          underlying ML library to produce a prediction.  The mapping back to
-  *                          (K, Double) pairs is also its responsibility.
+  *                          (K, Double) pairs is also its responsibility.  If the predictor
+  *                          produced by predictorProducer is Closeable, it will be closed when
+  *                          MultilabelModel's close method is called.
   * @param numMissingThreshold if provided, we check whether the threshold is exceeded.  If so,
   *                            return an error instead of the computed score.  This is for missing
   *                            data situations.
--- a/aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
+++ b/aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
@@ -85,6 +85,9 @@ extends SubmodelBase[U, Map[K, Double], A, B]
   @transient private[this] lazy val defaultLabelInfo =
     LabelsAndInfo(labelsInTrainingSet.indices, labelsInTrainingSet, Seq.empty, None)
 
+  /**
+    * Making from label to index into the sequence of all labels encountered during training.
+    */
   private[this] val labelToInd: Map[K, Int] =
     labelsInTrainingSet.zipWithIndex.map { case (label, i) => label -> i }(collection.breakOut)
 
--- a/aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
+++ b/aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
@@ -1,6 +1,6 @@
 package com.eharmony.aloha.models.multilabel
 
-import java.io.Closeable
+import java.io.{Closeable, PrintWriter, StringWriter}
 
 import com.eharmony.aloha.audit.Auditor
 import com.eharmony.aloha.dataset.density.Sparse
@@ -29,7 +29,7 @@ import scala.util.{Failure, Success, Try}
   *
   * Created by ryan.deak on 8/29/17.
   *
-  * @param modelId An identifier for the model.  User in score and error reporting.
+  * @param modelId An identifier for the model.  Used in score and error reporting.
   * @param featureNames feature names (parallel to featureFunctions)
   * @param featureFunctions feature extracting functions.
   * @param labelsInTrainingSet a sequence of all labels encountered during training. Note: the
@@ -78,11 +78,18 @@ extends SubmodelBase[U, Map[K, Double], A, B]
   @transient private[this] lazy val predictor = predictorProducer()
   predictor // Force predictor eagerly
 
+  /**
+    * Cache this in case labelsOfInterest is None.  In that case, we don't want to repeatedly
+    * create this because it could create a GC burden for no real reason.
+    */
+  @transient private[this] lazy val defaultLabelInfo =
+    LabelsAndInfo(labelsInTrainingSet.indices, labelsInTrainingSet, Seq.empty, None)
+
   private[this] val labelToInd: Map[K, Int] =
     labelsInTrainingSet.zipWithIndex.map { case (label, i) => label -> i }(collection.breakOut)
 
   override def subvalue(a: A): Subvalue[B, Map[K, Double]] = {
-    val li = labelsAndInfo(a, labelsInTrainingSet, labelsOfInterest, labelToInd)
+    val li = labelsAndInfo(a, labelsOfInterest, labelToInd, defaultLabelInfo)
 
     if (li.labels.isEmpty)
       reportNoPrediction(modelId, li, auditor)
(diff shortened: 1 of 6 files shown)


aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,334 @@
...
+      if (unsorted.size == labelsShouldPredict.size) Seq.empty
+      else labelsShouldPredict.filterNot(labelToInd.contains)
+
+    val (ind, lab) = unsorted.sortBy{ case (i, _) => i }.unzip
+
+    LabelsAndInfo(ind, lab, noPrediction, problems)
+  }
+
+
+  override def parser: ModelParser = Parser
```
deaktator, line 350: Please disregard the `Parser` **object** as it is just a stub for this phase of development.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/package.scala
```diff
@@ -0,0 +1,69 @@
...
+    * produces a sparse multi-label predictor.
+    *
+    * This definition is "lazy" because we can't guarantee that the underlying predictor is
+    * `Serializable` so we pass around a function that can be cached in a ''transient''
+    * `lazy val`.  This function should however be `Serializable` and testing should be done
+    * to ensure that each predictor producer is `Serializable`.
+    *
+    * @tparam K the type of labels (or classes in the machine learning literature).
+    */
+  type SparsePredictorProducer[K] = () => SparseMultiLabelPredictor[K]
```
deaktator, line 72: This is the only `type` publicly available.

deaktator: Not true anymore. Made them all package private except the last one so that they can be used for testing.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
...
+    * We don't care about the lazy property.  It should be created eagerly.
+    */
+  @transient private[this] lazy val predictor = predictorProducer()
+  predictor // Force predictor eagerly
+
+  /**
+    * Cache this in case labelsOfInterest is None.  In that case, we don't want to repeatedly
+    * create this because it could create a GC burden for no real reason.
+    */
+  @transient private[this] lazy val defaultLabelInfo =
```
deaktator: Probably does not need to be a lazy val. Consider changing.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
...
+    LabelsAndInfo(labelsInTrainingSet.indices, labelsInTrainingSet, Seq.empty, None)
+
+  /**
+    * Making from label to index into the sequence of all labels encountered during training.
+    */
+  private[this] val labelToInd: Map[K, Int] =
+    labelsInTrainingSet.zipWithIndex.map { case (label, i) => label -> i }(collection.breakOut)
+
+  override def subvalue(a: A): Subvalue[B, Map[K, Double]] = {
```
deaktator, line 99: Is this easily understandable? I tried to make it as understandable and testable a possible.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
...
+
+        predictionTry match {
+          case Success(pred) => reportSuccess(modelId, li, missing, pred, auditor)
+          case Failure(ex)   => reportPredictorError(modelId, li, missing, ex, auditor)
+        }
+      }
+    }
+  }
+
+  override def close(): Unit =
```
deaktator: Probably add a comment about what is closed and when.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
...
+    val labelsShouldPredict = labelsOfInterest(example)
+
+    val unsorted =
+      for {
+        label <- labelsShouldPredict
+        ind   <- labelToInd.get(label).toList
+      } yield (ind, label)
+
+    val problems =
+      if (labelsShouldPredict.nonEmpty) None
```
deaktator, line 337: Is this a good check? Should we be more strict about when to report errors?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
...
+  *                          underlying ML library to produce a prediction.  The mapping back to
+  *                          (K, Double) pairs is also its responsibility.  If the predictor
+  *                          produced by predictorProducer is Closeable, it will be closed when
+  *                          MultilabelModel's close method is called.
+  * @param numMissingThreshold if provided, we check whether the threshold is exceeded.  If so,
+  *                            return an error instead of the computed score.  This is for missing
+  *                            data situations.
+  * @param auditor transforms a `Map[K, Double]` to a `B`.  Reports successes and errors.
+  * @tparam U upper bound on model output type `B`
+  * @tparam K type of label or class
```
deaktator: `K` is not constrained.  Should we make the constructor private and then have two factory methods: one where `K <: AnyVal` and one where `K <: Serializable`?  They could both thunk down to the unconstrained constructor.  What do you think about this?  If `K` is constrained, we can **guarantee** the serializability of `MultilabelModel`.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
+package com.eharmony.aloha.models.multilabel
```
jon-morra-zefr, line 1: Why bother to have a multilabel package that only has one class?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
...
+  *
+  * Created by ryan.deak on 8/29/17.
+  *
+  * @param modelId An identifier for the model.  Used in score and error reporting.
+  * @param featureNames feature names (parallel to featureFunctions)
+  * @param featureFunctions feature extracting functions.
+  * @param labelsInTrainingSet a sequence of all labels encountered during training. Note: the
+  *                            order of labels may relate to the predictor produced by
+  *                            predictorProducer.  It is the caller's responsibility to ensure
+  *                            the order is correct.  To mitigate such problems, both labels
```
jon-morra-zefr, line 39: What is a `correct order`?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
...
+  *                            and indices into labelsInTrainingSet are passed to the predictor
+  *                            produced by predictorProducer.
+  * @param labelsOfInterest if provided, a sequence of labels will be extracted from the example
+  *                         for which a prediction is desired. The ''intersection'' of the
+  *                         extracted labels and the training labels will be the labels for which
+  *                         predictions will be produced.
```
jon-morra-zefr, line 45: I think we should add something like `If a provided label is not in the training label set it is ignored`.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
jon-morra-zefr: Yes, for sure we should do this.  This class has to be serializable.  If that's not a test, please add it.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
...
+   with RegressionFeatures[A] {
+
+  import MultilabelModel._
+
+  /**
+    * predictory is transient lazy value because we don't need to worry about serialization.
+    * We don't care about the lazy property.  It should be created eagerly.
+    */
+  @transient private[this] lazy val predictor = predictorProducer()
+  predictor // Force predictor eagerly
```
jon-morra-zefr, line 82: Why bother with `lazy` here?  Wouldn't this be the same as the above two lines

`@transient private[this] val predictor = predictorProducer()`

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
jon-morra-zefr: Agree

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
jon-morra-zefr: This looks good to me.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
jon-morra-zefr: I don't think it's necessary.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
...
+      problems: Option[GenAggFuncAccessorProblems]
+  ) {
+    def missingVarNames: Seq[String] = problems.map(p => p.missing).getOrElse(Nil)
+    def errorMsgs: Seq[String] = {
+      missingLabels.map { lab => s"Label not in training labels: $lab" } ++
+      problems.map(p => p.errors).getOrElse(Nil)
+    }
+  }
+
+  private[multilabel] val NumLinesToKeepInStackTrace = 20
```
jon-morra-zefr, line 161: This seems like a variable that is more general than just this class.  Is that so?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
...
+
+  private[multilabel] val TooManyMissingError =
```
jon-morra-zefr, line 163: Same here.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,387 @@
...
+    * @return a SubValue indicating failure.
+    */
+  protected[multilabel] def reportTooManyMissing[U, K, B <: U](
+      modelId: ModelIdentity,
+      labelInfo: LabelsAndInfo[K],
+      missing: scm.Map[String, Seq[String]],
+      auditor: Auditor[U, Map[K, Double], B]
+  ): Subvalue[B, Nothing] = {
+
+    // TODO: Check that missing.values.flatten.toSet AND labelInfo.missingFeatures have the same format.
```
jon-morra-zefr, line 217: Is this TODO still valid?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
jon-morra-zefr: I think this seems sufficient.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/package.scala
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+  * Created by ryan.deak on 8/31/17.
+  */
+package object multilabel {
+
+  // All but the last type are package private, for testing.  The last is public.
+
+  /**
+    * Features about the input value (NOT including features based on labels).
```
jon-morra-zefr: I agree with this, how hard is this to migrate?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/package.scala
```diff
@@ -0,0 +1,74 @@
+package com.eharmony.aloha.models
```
jon-morra-zefr, line 1: Most things here should live at a higher level IMO.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -0,0 +1,170 @@
...
+  private val Auditor = RootedTreeAuditor.noUpperBound[Map[Label, Double]]()
+
+// TODO: Access information returned in audited value by using the following functions:
+  //    val aud: RootedTree[Any, Map[Label, Double]] = ???
+  //    aud.modelId           // : ModelIdentity
+  //    aud.value             // : Option[Map[Label, Double]]  // Should be missing on failure.
+  //    aud.missingVarNames   // : Set[String]
+  //    aud.errorMsgs         // : Seq[String]
+  //    aud.prob              // : Option[Float]  (Shouldn't need this)
+}
```
jon-morra-zefr, line 170: newline, consider using scalafmt

jon-morra-zefr (changes requested):
Left some comments to consider.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: Has to be lazy.

```scala
import java.io.{
  ByteArrayInputStream, 
  ByteArrayOutputStream, 
  ObjectInputStream,
  ObjectOutputStream}

case class Fail(i: Int) {
  @transient val s = i.toString
}

case class Success(i: Int) {
  @transient lazy val s = i.toString
}

def serializeRoundTrip[A <: AnyRef](a: A) = {
  val baos = new ByteArrayOutputStream
  val oos = new ObjectOutputStream(baos)
  oos.writeObject(a)
  oos.close
  val ois = new ObjectInputStream(new ByteArrayInputStream(baos.toByteArray))
  ois.readObject().asInstanceOf[A]
}

val fail = serializeRoundTrip(Fail(1))
assert(fail.s == null)

val success = serializeRoundTrip(Success(1))
assert(success.s == "1")
```

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: Did it anyway.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: OK

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: I added a type class that can ensure types are `Serializable` or `AnyVal`.  I like this approach better than making the constructor private and adding multiple factory methods.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: To give the `multilabel` package object somewhere to live.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: Yeah.  But I don't really want to aggregate this right now.  I don't know exactly where it should go.  I need to think more about how to do this globally.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: Yeah.  I'm going to let Amir test this.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/package.scala
deaktator: That's why they are package private.  They are only there to provide named building blocks to make the final type.

deaktator: I wanted to make the definitions as self contained as possible.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
deaktator: I will in the future.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/package.scala
deaktator: The `sci.IndexedSeq`?  It's not efficient enough when I tested a long time ago.

jmorra (approved):

deaktator:
OK.  Dismissing this PR and moving on to stage 2: The companion object.

PR #179, opened 2017-09-08
Close #163 parsing

### LINKED ISSUE #163: Support multilabel problems

deaktator:
## Stage 2 Model Parsing 

In this PR, reviewers should concentrate on `MultilabelModel` parsing.  This is a beast because we want a `JsonReader` if `N` is `Map[K, Double]` but we don't have the information at the type level to ensure that.  So conservatively only create a parser if `N` is a `Map[_, Double]` and N has two type arguments.  This is not totally safe but should hopefully suffice.  Consequently, there is a lot of type hackery (casting) in making the two types isomorphic even though the the compiler can't prove that they are.

This includes the `Parser` in the `MultilabelModel`*companion object* as well as:

### *multilabel* package 
* `MultilabelModelParserPlugin.scala`
* `MultilabelPluginProviderCompanion.scala`
* `PluginInfo.scala`

### *multilabel/json* package 
* `MultilabelModelJson.scala`
* `MultilabelModelReader.scala`


jon-morra-zefr:
Should this build be passing?  If not let's get it passing before reviewing.

deaktator:
@jmorra. Doesn't pass because tests don't succeed. Compiles. Amir is working on tests.

deaktator:
will merge to `development` branch.

PR #182, opened 2017-09-23
163 multilabel test

amirziai:
@deaktator


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
+### Summary
```
deaktator: I like this idea and want to have it.  Can you to delete this file, create an issue, add it back and issue a separate PR directly into `master`.  I would like to track it separately.

aloha-core/src/main/scala/com/eharmony/aloha/semantics/SemanticsUdfException.scala
```diff
@@ -26,7 +26,7 @@ case class SemanticsUdfException[+A](
         input: A)
     extends AlohaException(SemanticsUdfException.getMessage(specification, accessorOutput, accessorsMissingOutput, accessorsInErr, input), cause)
 
-private object SemanticsUdfException {
```
deaktator, line 29: *Good catch us!* very nice.

aloha-core/src/main/scala/com/eharmony/aloha/util/SerializabilityEvidence.scala
```diff
@@ -10,8 +10,6 @@ object SerializabilityEvidence {
 
   implicit def anyValEvidence[A <: AnyVal]: SerializabilityEvidence[A] =
```
deaktator: Can you add these two delete spaces back in.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -1,44 +1,60 @@
...
 
   @Test def testModelCloseClosesPredictor(): Unit = {
-    // Make the predictorProducer passed to the constructor be a
-    //   'SparsePredictorProducer[K] with Closeable'.
-    // predictorProducer should track whether it is closed (using an AtomicBoolean or something).
-    // Call close on the MultilabelModel instance and ensure that the underlying predictor is
-    // also closed.
-
-    fail()
+    case class PredictorClosable[K](prediction: Double = 0d)
```
deaktator: `(prediction: Double = 0d)` looks unnecessary.  You could make this a regular class with no argument list.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -1,44 +1,60 @@
...
+    val model = modelNoFeatures.copy(predictorProducer = Lazy(predictor))
+
+    model.close()
+    assertTrue(predictor.isClosed)
   }
 
   @Test def testLabelsOfInterestOmitted(): Unit = {
-    // Test labelsAndInfo[A, K] function.
-    //
     // When labelsOfInterest = None, labelsAndInfo should return:
```
deaktator: You can probably remove this commented out code.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,124 +63,336 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
   @Test def testLabelsOfInterestProvided(): Unit = {
-    // Test labelsAndInfo[A, K] function.
-    //
-    // labelsAndInfo(a, labelsInTrainingSet, labelsOfInterest, labelToInd) ==
-    // labelsForPrediction(a, labelsOfInterest.get, labelToInd)
-
-    fail()
+    val a = ()
+    val labelsOfInterest =
+      Option(GenFunc0[Unit,sci.IndexedSeq[Label]]("", _ => labelsInTrainingSet))
```
deaktator: Move this option to:

```scala
labelsOfInterest = Option(labelsOfInterest),
```

Then you won't have to call `.get` in `labelsOfInterest.get`.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,124 +63,336 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
+      auditor    = Auditor
+    )
+    assertEquals(Some(predictions), report.natural)
+    assertEquals(Some(predictions), report.audited.value)
+    assertEquals(report.natural, report.audited.value)
+  }
+
+  @Test def testLabelsForPredictionContainsProblemsWhenNoLabelProvided(): Unit = {
+    val labelsAndInfoNoLabels = labelsForPrediction(
+      example          = Map[Label, Label](),  // no label provided
```
deaktator: Should both of these type parameters really be `Label`? Or is it the case that one or both are just `String`s?  Are you "*abusing notation*" here?

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,124 +63,336 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
   }
 
-  @Test def testLabelsForPredictionContainsProblemsWhenLabelsIsEmpty(): Unit = {
-    // Test this:
-    //    val problems =
-    //      if (labelsShouldPredict.nonEmpty) None
-    //      else Option(labelsOfInterest.accessorOutputProblems(a))
+  @Test def testLabelsForPredictionContainsProblemsWhenLabelsIsNotPresent(): Unit = {
+    val labelsExtractor =
+      (label: Map[Label, Label]) => label.get("labels") match {
```
deaktator: This can't be right.  Look at the types.  Looks like you want `labelsExtractor` to have the following type:

```scala
val labelsExtractor: Map[String, String] => Option[sci.IndexedSeq[String]] = ???
```

But then, think about what you're trying to do. `label.get("labels")` must return a `Option[String]`, based on the API.  So, you are checking if an `Option[String]` is a `sci.IndexedSeq[_]`.  Then if it is, you're wrapping it in another Option.  I don't understand.  First of all:

```scala
Option.empty[String].isInstanceOf[sci.IndexedSeq[_]]   // false
Option("").isInstanceOf[sci.IndexedSeq[_]]             // false
```

It looks like you are trying to subvert the type system but I don't see the benefit.  It's in instances like this that the compiler warns about erasure when you do `case ls: sci.IndexedSeq[_] =>` because this is probably a bad check.  This should probably be something like the following instead: 

```scala
(label: Map[Label, Label]) => label.get("labels") match {
  // strVal is a string.  Do something with it to create a Option[sci.IndexedSeq[String]].
  case Some(strVal) => ??? 
  case None => None
}
```

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,124 +63,336 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
+    val descriptor = "label accessor that throws"
+    val labelsOfInterest =
+      GenFunc.f1(GeneratedAccessor(descriptor,
+        ( _ => throw new Exception()) : Map[String, String] => Option[sci.IndexedSeq[String]], None))(
+        "", _ getOrElse sci.IndexedSeq.empty[Label]
+      )
+    val labelsOfInterestWrapped = EnrichedErrorGenAggFunc(labelsOfInterest)
+
+    val problemsNoLabelsExpected = Option(GenAggFuncAccessorProblems(Seq(), Seq(descriptor)))
+    Try(
```
deaktator: Can you just do something like what follows?  It seems like it would be cleaner.

```scala
Try { 
  labelsForPrediction(Map.empty, labelsOfInterestWrapped, labelsInTrainingSetToIndex)
} match {
  // Failure is scala.util.Failure
  case Failure(SemanticsUdfException(_, _, _, accessorsInErr, _, _)) => 
    assertEquals(accessorsInErr, problemsNoLabelsExpected.get.errors)
  case _ => fail()
}
```

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,124 +63,336 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
-    //      else labelsShouldPredict.filterNot(labelToInd.contains)
-
-    fail()
+    val labelNotInTrainingSet = "d"
+    val labelsAndInfo = labelsForPrediction(Map("label" -> labelNotInTrainingSet),
+      labelsOfInterestExtractor, labelsInTrainingSetToIndex)
+    assertEquals(Seq(labelNotInTrainingSet), labelsAndInfo.labelsNotInTrainingSet)
   }
 
   @Test def testLabelsForPredictionReturnsLabelsSortedByIndex(): Unit = {
```
deaktator: I would put a comment here that this test would likely be done better in a property-based testing framework like ScalaCheck.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,124 +63,336 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
+      SparseMultiLabelPredictor[Label] {
+      override def apply(v1: SparseFeatures,
+        v2: Labels[Label],
+        v3: LabelIndices,
+        v4: SparseLabelDepFeatures): Try[Map[Label, Double]] = Try(throw new Exception("error"))
+    }
+
+    val modelWithThrowingPredictorProducer = modelWithFeatureFunctions.copy(
+      predictorProducer = Lazy(PredictorThatThrows),
+      labelsOfInterest    = None
```
deaktator: spacing is weird.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,124 +63,336 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
+
+    val result = Try(modelWithFeatureFunctionThatThrows(Map()))
+    result match {
+      case Success(_) => fail()
+      case Failure(ex) => assertEquals(exception, ex)
+    }
   }
 }
 
 object MultilabelModelTest {
```
deaktator, line 295: Everything in here should be `private` scoped I think.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,124 +63,336 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
+      ldfUnused: SparseLabelDepFeatures): Try[Map[K, Double]] =
+        Try(labels.map(_ -> prediction).toMap)
+  }
+
+  case class Lazy[A](value: A) extends (() => A) {
+    override def apply(): A = value
+  }
+
+  // Common input
+  val labelsInTrainingSet: sci.IndexedSeq[Label] = sci.IndexedSeq[Label]("a", "b", "c")
```
deaktator: These constants [should be capitalized](http://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods).

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,124 +63,336 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
+    featureFunctions    = sci.IndexedSeq[GenAggFunc[Vector[String], Sparse]](),
+    labelsInTrainingSet = labelsInTrainingSet,
+    labelsOfInterest    = Some(GenFunc0("", (a: Vector[String]) => a))
+  )
+
+  // LabelsAndInfo
+  val labelsAndInfoEmpty = LabelsAndInfo(
+    indices = sci.IndexedSeq.empty,
+    labels = sci.IndexedSeq.empty,
+    labelsNotInTrainingSet = Seq[Label](),
```
deaktator: Just do `Seq.empty[Label]`.  See [GenericCompanion.scala](https://github.com/scala/scala/blob/v2.11.8/src/library/scala/collection/generic/GenericCompanion.scala#L38-L46) for more info.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,124 +63,336 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
+    problems = None
+  )
+  val labelsAndInfoMissingLabels: LabelsAndInfo[Label] =
+    labelsAndInfoEmpty.copy(labelsNotInTrainingSet = labelsNotInTrainingSet)
+
+  // Reports
+  val reportNoPredictionPartial:
+    (LabelsAndInfo[Label]) => Subvalue[RootedTree[Any, Map[Label, Double]], Nothing] =
+    reportNoPrediction(
```
deaktator: indenting is bad here.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,124 +63,336 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
+    modelId   = ModelId(),
+    _: LabelsAndInfo[Label],
+    auditor   = Auditor
+  )
+  val reportNoPredictionEmpty: Subvalue[RootedTree[Any, Map[Label, Double]], Nothing] =
+    reportNoPredictionPartial(labelsAndInfoEmpty)
+
+  // Throwable and stack trace
+  def getThrowable(errorMessage: String): (Throwable, String) = {
+    val throwable = Try(throw new Exception(errorMessage)).failed.get
```
deaktator: I would add a comment here why you're doing this.  You might be able to change this to:

```scala
val throwable = new Exception(errorMessage)
```

Not sure though.

deaktator (changes requested):
Most of this stuff looks good.  Line 162 does not though.  I would definitely change that and look work on the other comments too.  Don't worry too much about the comment in `getThrowable`.  I just suggested that in case it also works because it's less code.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
deaktator: Which means that this could probably just be done with a `flatMap` (or `map`) after calling `label.get`.

aloha-core/src/main/scala/com/eharmony/aloha/util/SerializabilityEvidence.scala
```diff
@@ -13,5 +13,5 @@ object SerializabilityEvidence {
 
   implicit def serializableEvidence[A <: java.io.Serializable]: SerializabilityEvidence[A] =
     new SerializabilityEvidence[A]{}
-}
 
+}
```
deaktator, line 17: Can you add a newline at the end of the file.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -47,51 +46,42 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
...
 
   @Test def testLabelsOfInterestProvided(): Unit = {
     val a = ()
-    val labelsOfInterest =
-      Option(GenFunc0[Unit,sci.IndexedSeq[Label]]("", _ => labelsInTrainingSet))
+    val labelsOfInterest = GenFunc0[Unit,sci.IndexedSeq[Label]]("", _ => LabelsInTrainingSet)
     val actual: LabelsAndInfo[Label] = labelsAndInfo(
       a = a,
-      labelsOfInterest = labelsOfInterest,
+      labelsOfInterest = Option(labelsOfInterest),
```
deaktator: Much better.

aloha-core/src/test/scala/com/eharmony/aloha/models/multilabel/MultilabelModelTest.scala
```diff
@@ -159,21 +149,16 @@ class MultilabelModelTest extends ModelSerializationTestHelper {
 
   @Test def testLabelsForPredictionContainsProblemsWhenLabelsIsNotPresent(): Unit = {
     val labelsExtractor =
-      (label: Map[Label, Label]) => label.get("labels") match {
-        case ls: sci.IndexedSeq[_] => Option(ls.asInstanceOf[sci.IndexedSeq[Label]])
-        case _ => None
-      }
-
+      (label: Map[Label, Label]) => label.get("labels").map(sci.IndexedSeq[Label](_))
```
deaktator: This looks way better.  Did it make sense why?

deaktator (approved):
Much better.  Looks good.

aloha-core/src/main/scala/com/eharmony/aloha/util/SerializabilityEvidence.scala
amirziai: I think it has one

PR #188, opened 2017-10-03
Closes #185

### LINKED ISSUE #185: Make an Either Auditor
The left will be for failure messages.  The right will be for successes.  The `Right` values will not track compromises made and diagnostic errors.  Those will only be captured in the case of failures.

deaktator:
## Summary
Provides an auditor that returns an Either.  The Either has failure information or a success value.

There is some commented code in the check-in.  That is there on purpose because I wanted to get opinions about wether to include the code that was commented out or whether it should be deleted.

## Resolves
* Fixes issue #185 

## Bug Fixes/New Features
* Either Auditor

## Side Effects
`None` (*ha ha*)

## How to Verify
Check tests

## Tests
`EitherAuditorTest` should be comprehensive.

### Code Reviewer(s)
@jmorra


deaktator:
BTW.  I'm leaning towards deleting the commented out code rather than including the code that is commented out.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
```diff
@@ -0,0 +1,175 @@
...
+  * a "''successful''" prediction is produced, even if compromises were made in order to
+  * produce the prediction or problems like missing data were encountered.
+  *
+  * To construct an [[EitherAuditor]], use the `apply` factory methods in the companion
+  * object.
+  *
+  * Created by ryan.deak on 10/2/17.
+  * @tparam A the type of value returned in the success case.
+  */
+sealed abstract class EitherAuditor[A]
```
deaktator, line 22: No constructor exposed.  Only factory.  That way we can control the creation of the two subclasses.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
```diff
@@ -0,0 +1,175 @@
...
+  override def changeType[M: RefInfo]: Some[EitherAuditor[M]] = Some(changeTo[M])
+
+  override def failure(
+      modelId: ModelIdentity,
+      errorMsgs: => Seq[String] = Nil,
+      missingVarNames: => Set[String] = Set.empty,
+      subvalues: Seq[Either[EitherAuditorError, _]] = Nil
+  ): Either[EitherAuditorError, A] = {
+    val (errors, missing, failIds) = aggregateDiagnostics(errorMsgs, missingVarNames, subvalues)
+    Left(EitherAuditorError(::(modelId, failIds), errors, missing))
```
deaktator, line 39: Notice since modelId is always inserted at every level, even if it's filtered out later, the list of model IDs is always non-empty, hence encoded in the type.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
```diff
@@ -0,0 +1,175 @@
...
+      subvalues: Seq[Either[EitherAuditorError, _]] = Nil,
+      prob: => Option[Float] = None
+  ): Either[EitherAuditorError, A] = Right(valueToAudit)
+
+  /**
+    * Construct a new EitherAuditor[M].  Notice this allows succeeds.
+    * @tparam M the type of success value in the new auditor.
+    * @return
+    */
+  protected[this] def changeTo[M]: EitherAuditor[M]
```
deaktator, line 56: Added this because an auditor can always be produced.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
```diff
@@ -0,0 +1,175 @@
...
+  */
+sealed abstract class EitherAuditor[A]
+extends MorphableAuditor[Either[EitherAuditorError, _], A, Either[EitherAuditorError, A]] {
+
+  /**
+    * Return an EitherAuditor for success type `M`.
+    * @tparam M type of values returned on success
+    * @return a new auditor. An EitherAuditor can be returned for any type `M`.
+    */
+  override def changeType[M: RefInfo]: Some[EitherAuditor[M]] = Some(changeTo[M])
```
deaktator, line 30: Notice return type `Some`.  I think this is a good idea since function codomains are covariant.  Any downside?

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  override def success(
+      modelId: ModelIdentity,
+      valueToAudit: A,
+      errorMsgs: => Seq[String] = Nil,
+      missingVarNames: => Set[String] = Set.empty,
+      subvalues: Seq[Either[EitherAuditorError, _]] = Nil,
+      prob: => Option[Float] = None
+  ): Either[EitherAuditorError, A] = Right(valueToAudit)
```
deaktator, line 49: Just `Right(valueToAudit)`.  Beautiful.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
```diff
@@ -0,0 +1,175 @@
...
+}
+
+object EitherAuditor {
+
+  /**
+    * Produce an [[EitherAuditor]] that aggregates diagnostic information.
+    * @tparam A the type of value returned in the success case.
+    * @return
+    */
+  def apply[A]: EitherAuditor[A] = apply[A](aggregateDiagnostics = true)
```
deaktator, line 81: I like this better than a default parameter.  More binary combat (If we care) (TODO: add MIMA)

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
```diff
@@ -0,0 +1,175 @@
...
+  * `EitherAuditor[A](aggregateDiagnostics = false)`.
+  *
+  * If the auditor is created via `EitherAuditor[A](aggregateDiagnostics = true)`,
+  * or simply `EitherAuditor[A]`, then the trees are pruned at successes.  For
+  * instance, imagine a model, model `1`, that has submodels `2` and `5`.
+  * Submodel `2` has submodels `3` and `4`; submodel `5` has submodels `6` and `7`.
+  * Let's say the following submodels fail: 1, 2, 4, 6, 7.  Then, information would
+  * be aggregated as follows:
+  *
+  * {{{
```
deaktator, line 138: Lovely ASCII art!

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
```diff
@@ -0,0 +1,175 @@
...
+    override protected[this] def aggregateDiagnostics(
+        err: Seq[String],
+        missing: Set[String],
+        sub: Seq[Either[EitherAuditorError, _]]): (Seq[String], Set[String], List[ModelIdentity]) =
+      (err, missing, Nil)
+  }
+}
+
+/**
+  * Diagnostic information about failures encountered.  This structure can be thought
```
deaktator, line 125: Does this comment make sense?

aloha-core/src/test/scala/com/eharmony/aloha/audit/impl/EitherAuditorTest.scala
```diff
@@ -0,0 +1,299 @@
...
+    val v7 = a.failure(ModelId(7))
+    val v5 = a.failure(ModelId(5), subvalues = Seq(v6, v7))
+
+    val v1 = a.failure(ModelId(1), subvalues = Seq(v2, v5))
+
+    val mids = v1.left.get.failureModelIds
+    assertEquals(1 to 7, mids.map(mid => mid.getId()))
+  }
+
+  @Test def testScaladocExample(): Unit = {
```
deaktator, line 239: Proof the documentation is accurate.

deaktator (commented):
What does everyone else think?  @jmorra?

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
jmorra: Just looking at the signature of this function it looks wonky.  Why bother with a `Some()` if failure is impossible.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
```diff
@@ -0,0 +1,175 @@
...
+      AggEitherAuditor[A]()
+    else NoAggEitherAuditor[A]()
+
+  private[this] final case class AggEitherAuditor[A]() extends EitherAuditor[A] {
+    override protected[this] def changeTo[M]: EitherAuditor[M] = AggEitherAuditor[M]()
+    override protected[this] def aggregateDiagnostics(
+        err: Seq[String],
+        missing: Set[String],
+        sub: Seq[Either[EitherAuditorError, _]]): (Seq[String], Set[String], List[ModelIdentity]) = {
+      val e = sub.foldLeft(err)((s, sv) => s ++ sv.fold(e => e.errorMsgs, _ => Nil))
```
deaktator, line 102: These can't be `flatMap`.  Must be `foldLeft`.  Otherwise, there are type issues.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
jmorra: +1

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
deaktator, line 102: These are probably bad names.  Maybe:

```scala
val e = sub.foldLeft(err)((es, sv) => es ++ sv.fold(e => e.errorMsgs, _ => Nil))
val m = sub.foldLeft(missing)((ms, sv) => ms ++ sv.fold(e => e.missingVarNames, _ => Set.empty))
```

Do we care?

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
jmorra: I get it, `WartRemover` is on your mind.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
```diff
@@ -0,0 +1,175 @@
...
+
+  /**
+    * Produce an [[EitherAuditor]].
+    * @param aggregateDiagnostics  If `true`, aggregate diagnostics; otherwise,
+    *                              disregard submodel information in the
+    *                              [[EitherAuditorError]] returned by the auditor.
+    * @tparam A the type of value returned in the success case.
+    * @return
+    */
+  def apply[A](aggregateDiagnostics: Boolean): EitherAuditor[A] =
```
jmorra, line 91: Note that `WartRemover` wouldn't like you using the same function name (although I do ;) )

aloha-core/src/test/scala/com/eharmony/aloha/audit/impl/EitherAuditorTest.scala
```diff
@@ -0,0 +1,299 @@
...
+    //
+    //   1:F +-- 2:F +--  3:S
+    //       |       |
+    //       |       +--  4:F
+    //       |
+    //       +-- 5:S +--  6:F
+    //               |
+    //               +--  7:F
+    //
+    val v3 = a.success(ModelId(3), 3, Seq("e3"), Set("m3"))
```
deaktator, line 252: Should these be called: `f1`, `f2`, `s3`, `f4`, `s5`, `f6`, `f7`?  Is that any better?

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
```diff
@@ -0,0 +1,175 @@
...
+        missing: Set[String],
+        sub: Seq[Either[EitherAuditorError, _]]): (Seq[String], Set[String], List[ModelIdentity]) =
+      (err, missing, Nil)
+  }
+}
+
+/**
+  * Diagnostic information about failures encountered.  This structure can be thought
+  * of as a flattened tree of failures.  The tree is pruned one of two ways, it is
+  * either prune at the root, or pruned at the the first submodel successes
```
jmorra, line 127: prune => pruned

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
jmorra: Lovely.

aloha-core/src/test/scala/com/eharmony/aloha/audit/impl/EitherAuditorTest.scala
jmorra: Bonus points

jmorra (approved):
Overall, looks good.  Left a few comments.

aloha-core/src/main/scala/com/eharmony/aloha/audit/impl/EitherAuditor.scala
deaktator: The contract of `MorphableAuditor` is 

```scala
def changeType[M: RefInfo]: Option[ /* Auditor of similar type parametrized by M */ ]
```

It's possible that an auditor can't produce a similar auditor based on type `M`.  So we return an `Option`.  Because functions can return subtypes and `Some` is an `Option`, we can explicitly return `Some` since functions are covariant in their output type.

deaktator:
Merged to `development` branch.

PR #192, opened 2017-10-31
Closes #163

### LINKED ISSUE #163: Support multilabel problems

deaktator:
## Summary
Adds multi-label support for VW 

## Resolves
* Fixes #163

## Bug Fixes/New Features
* Automatic VW argument augmentation

## Tests
- `aloha-core/com.eharmony.aloha.dataset.vw.multilabel.VwMultilabelRowCreatorTest`
- `aloha-vw-jni/com.eharmony.aloha.models.vw.jni.multilabel`

### Code Reviewer(s)
@jmorra


deaktator:
Closed.  This should not be merged directly to master.  It should be merged to `development` then squash merged to master.

PR #193, opened 2017-11-01
Closes #163

### LINKED ISSUE #163: Support multilabel problems

deaktator:
## Summary
Adds multi-label model, a VW plugin for the multi-label model, a row creator for multi-label VW input.  Adds and `EitherAuditor` and fixes stack overflows in `tut` when generating documentation.

## Resolves
* Fixes #163

## How to Verify
Look at the tests.  I would also build locally (`sbt clean test package makeMicrosite`) to make sure nothing is left out.

## Tests
- `aloha-core/com.eharmony.aloha.dataset.vw.multilabel`
- `aloha-vw-jni/com.eharmony.aloha.models.vw.jni.multilabel`

## Code Reviewer(s)
@jmorra


aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+    s"$NegDummyClassId:$NegativeCost |$dummyClassNs $NegativeDummyClassFeature"
+
+  private[this] val positiveDummyStr =
+    s"$PosDummyClassId:$PositiveCost |$dummyClassNs $PositiveDummyClassFeature"
+
+  override def apply(a: A): (MissingAndErroneousFeatureInfo, Array[String]) = {
+    val (missingAndErrs, features) = featuresFunction(a)
+
+    // TODO: Should this be sci.BitSet?
+    val positiveIndices: Set[Int] =
```
jmorra: You might want to consider `BitSet` but only if the performance improvement is drastic.  If we care about performance shouldn't we consider not changing collection types?  Isn't `.toSeq` an O(N) operation?

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+      positiveDummyStr
+    )
+
+    (missingAndErrs, x)
+  }
+}
+
+object VwMultilabelRowCreator {
+
+  /**
```
jmorra: Love these descriptions.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+
+  private val NegativeDummyClassFeature = "N"
+
+  /**
+    * "shared" is a special keyword in VW multi-class (multi-row) format.
+    * See Hal Daume's [[https://www.umiacs.umd.edu/%7Ehal/tmp/multiclassVW.html page]].
+    *
+    * '''NOTE''': The trailing space should be here.
+    */
+  private[this] val SharedFeatureIndicator = "shared" + " "
```
jmorra: Why not just write this as `"shared "`?

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+    - "Λ", "λ"
+    *
+    * If one of these combinations cannot be used because at least one of the elements in a given
+    * row is in `usedNss`, then iterate over the Unicode set and take the first two characters
+    * found that adhere to are deemed a valid character.  These will then become the actual and
+    * dummy namespace names (respectively).
+    *
+    * The goal of this function is to try to use characters in literature used to denote a
+    * dependent variable.  If that isn't possible (because the characters are already used by some
+    * other namespace, just find the first possible characters.
```
jmorra: Close your parenthesis.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+
+  private[multilabel] def preferredLabelNamespaces(nss: sci.BitSet): Option[LabelNamespaces] = {
+    PreferredLabelNamespaces collectFirst {
+      case (actual, dummy) if !(nss contains actual.toInt) && !(nss contains dummy.toInt) =>
+        LabelNamespaces(actual, dummy)
+    }
+  }
+
+  private[multilabel] def nssToFirstCharBitSet(ss: Set[String]): sci.BitSet =
+    ss.collect { case s if s != "" =>
```
jmorra: `s.nonEmpty` is more legible

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+    * @param defaultNs the indices into `features` that should be placed in VW's default
+    *                  namespace.
+    * @param namespaces the indices into `features` that should be associated with each
+    *                   namespace.
+    * @param classNs a namespace for features associated with class labels
+    * @param dummyClassNs a namespace for features associated with dummy class labels
+    * @return an array to be passed directly to an underlying `VWActionScoresLearner`.
+    */
+  private[aloha] def trainingInput(
+      features: IndexedSeq[Sparse],
```
jmorra: Should these types be more general collections?

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+    val n = indices.size
+
+    // The length of the output array is n + 3.
+    //
+    // The first row is the shared features. These are features that are not label dependent.
+    // Then comes two dummy classes.  These are to make the probabilities work out.
+    // Then come the features for each of the n labels.
+    val x = new Array[String](n + 3)
+
+    val shared = VwRowCreator.unlabeledVwInput(features, defaultNs, namespaces, false)
```
jmorra: Can you name the boolean parameter.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+    x(1) = negativeDummyStr
+    x(2) = positiveDummyStr
+
+    // This is mutable because we want speed.
+    var i = 0
+    while (i < n) {
+      val labelInd = indices(i)
+
+      // TODO or positives.contains(labelInd)?
+      val dv = if (positiveLabelIndices(i)) PositiveCost else NegativeCost
```
jmorra, line 257: So the only problem with the way this is written now is that the function as provided doesn't guarantee an O(1) runtime which you must have for performance.  I would consider changing it so the reader of this function knows that it's an O(1) algorithm.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+  ): Array[String] = {
+
+    val n = indices.size
+
+    // Use a (mutable) array (and iteration) for speed.
+    // The first row is the shared features. These are features that are not label dependent.
+    // Then come the features for each of the n labels.
+    val x = new Array[String](n + 1)
+
+    val shared = VwRowCreator.unlabeledVwInput(features, defaultNs, namespaces, false)
```
jmorra, line 292: I see that this called function is tail recursive with the following comment from @deaktator 

```
// RMD 2015-06-12: GOD I HATE THIS CODE!!!  Maybe functionalize it in the future!
```

Since we're concerned about performance here, is this worth revisiting now?

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+         with RowCreatorProducerName
+         with VwCovariateProducer[A]
+         with DvProducer
+         with SparseCovariateProducer
+         with CompilerFailureMessages {
+
+    override type JsonType = VwMultilabeledJson
+
+    /**
+      * Attempt to parse the JSON AST to an intermediate representation that is used
```
jmorra, line 333: Can you finish your thought here.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+    private[multilabel] def positiveLabelsFn(
+        semantics: CompiledSemantics[A],
+        positiveLabels: String
+    ): Try[GenAggFunc[A, sci.IndexedSeq[K]]] =
+      getDv[A, sci.IndexedSeq[K]](
+        semantics, "positiveLabels", Option(positiveLabels), Option(Vector.empty[K]))
+  }
+
+  private[aloha] final case class LabelNamespaces(labelNs: Char, dummyLabelNs: Char)
+}
```
jmorra: newline

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
```diff
@@ -0,0 +1,361 @@
...
+final case class VwMultilabelRowCreator[-A, K](
+    allLabelsInTrainingSet: sci.IndexedSeq[K],
+    featuresFunction: FeatureExtractorFunction[A, Sparse],
+    defaultNamespace: List[Int],
+    namespaces: List[(String, List[Int])],
+    normalizer: Option[CharSequence => CharSequence],
+    positiveLabelsFunction: GenAggFunc[A, sci.IndexedSeq[K]],
+    classNs: Char,
+    dummyClassNs: Char,
+    includeZeroValues: Boolean = false
```
jmorra: I think we're missing something important here.  We have experimentally shown that when training on a very large multilabel problem the ability to randomly sample from the negative labels at train time both decreases train time and can increase model performance.  I think that a parameter to the row creator should be a function that allows me to sample a percentage of negative labels to add to each train line.  This needs to be random and the negative labels added needs to change with EVERY line.

aloha-core/src/main/scala/com/eharmony/aloha/factory/ri2jf/CollectionTypes.scala
```diff
@@ -1,16 +1,39 @@
...
+            k <- conv(tKey)
+            v <- conv(tVal)
+            f <- jf(mapFormat(k, v))
+          } yield f
+        case _ =>
+          // TODO: perhaps change the API at some point to better report errors here.
+          None
+      }
+    else if (RefInfoOps.isSubType[A, collection.immutable.Seq[Any]])
+//      conv(r.typeArguments.head).flatMap(f => jf(immSeqFormat(f)))
```
jmorra: remove dead code.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,495 @@
...
+
+import scala.collection.{immutable => sci, mutable => scm}
+import scala.util.{Failure, Success}
+
+
+// TODO: When adding label-dep features, a Seq[GenAggFunc[K, Sparse]] will be needed.
+// TODO: To create a Seq[GenAggFunc[K, Sparse]], a Semantics[K] needs to be derived from a Semantics[A].
+// TODO: MorphableSemantics provides this.  If K is *embedded inside* A, it should be possible in some cases.
+// TODO: An alternative is to pass a Map[K, Sparse], Map[K, Option[Sparse]], Map[K, Seq[Sparse]] or something.
+// TODO: Directly passing the map of LDFs avoids the need to derive a Semantics[K].  This is easier to code.
```
jmorra: I don't like the mention of LDF here.  That is a VW specific term and this class should not be VW specific.

jmorra: When we include TensorFlow support, for instance, it should hopefully be able to make use of this class.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,495 @@
...
+extends SubmodelBase[U, Map[K, Double], A, B]
+   with RegressionFeatures[A] {
+
+  import MultilabelModel._
+
+  /**
+    * predictory is transient lazy value because we don't need to worry about serialization.
+    * We don't care about the lazy property.  It should be created eagerly.
+    */
+  @transient private[this] lazy val predictor = predictorProducer()
```
jmorra, line 75: Have you tested this code within Spark?  Is it worth testing there to catch `Serializable` issues?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,495 @@
...
+  predictor // Force predictor eagerly
+
+  /**
+    * Cache this in case labelsOfInterest is None.  In that case, we don't want to repeatedly
+    * create this because it could create a GC burden for no real reason.
+    */
+  private[this] val defaultLabelInfo =
+    // Hopefully, making this non-transient doesn't increase the serialized size much.
+    // labelsInTrainingSet isn't serialized twice is it?
```
jmorra, line 84: Unless this is `lazy` I think the size will increase.  I do not know if that is a problem.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,495 @@
...
+    LabelsAndInfo(labelsInTrainingSet.indices, labelsInTrainingSet, Seq.empty, None)
+
+  /**
+    * Making from label to index into the sequence of all labels encountered during training.
+    */
+  private[this] val labelToInd: Map[K, Int] =
```
jmorra, line 90: See comment above.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,495 @@
...
+      case _ =>
+    }
+}
+
+object MultilabelModel extends ParserProviderCompanion {
+
+  /**
+    * Contains information about the labels to be used for predictions, and problems encountered
+    * while trying to get those labels.
+    * @param indices indices into the sequence of all labels seen during training.  These should
```
jmorra, line 131: If these are parallel arrays, then why not use an `IndexedSeq[(Int, K)]` to force the same length in the type?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
```diff
@@ -0,0 +1,495 @@
...
+    *                   labels seen in the training set.
+    * @tparam A input type of the model
+    * @tparam K type of label or class
+    * @return labels and information about the labels.
+    */
+  protected[multilabel] def labelsForPrediction[A, K](
+      example: A,
+      labelsOfInterest: GenAggFunc[A, sci.IndexedSeq[K]],
+      labelToInd: Map[K, Int]
+  ): LabelsAndInfo[K] = {
```
jmorra, line 325: How often is this function called?  It seems like it can be written more efficiently if it's hot.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModelParserPlugin.scala
```diff
@@ -0,0 +1,40 @@
+package com.eharmony.aloha.models.multilabel
+
+import com.eharmony.aloha.reflect.{RefInfo, RuntimeClasspathScanning}
+import spray.json.{JsonFormat, JsonReader}
+
+/**
+  * A plugin that will produce the
```
jmorra: Finish your thought.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModelParserPlugin.scala
```diff
@@ -0,0 +1,40 @@
...
+
+object MultilabelModelParserPlugin extends RuntimeClasspathScanning {
+
+  /**
+    * Finds the plugins in the `com.eharmony.aloha` namespace.
+    */
+  // TODO: Consider making this implicit so that we can parameterize the parser implicitly on a sequence of plugins.
+  protected[multilabel] def plugins(): Seq[MultilabelModelParserPlugin] =
+    scanObjects[MultilabelPluginProviderCompanion, MultilabelModelParserPlugin]("multilabelPlugin")
+}
```
jmorra, line 40: newline

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/PluginInfo.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import com.eharmony.aloha.models.reg.json.Spec
+
+import scala.collection.immutable.ListMap
+
+/**
+  * Created by ryan.deak on 9/7/17.
+  */
+trait PluginInfo[K] {
+  def features: ListMap[String, Spec]
```
jmorra, line 11: Why is this a `ListMap`?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/json/MultilabelModelJson.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import scala.collection.immutable.ListMap
+import com.eharmony.aloha.factory.ScalaJsonFormats
+
+trait MultilabelModelJson extends SpecJson with ScalaJsonFormats {
+
+  protected[this] case class Plugin(`type`: String)
+
+  /**
+    * Data for the
```
jmorra: Finish your thought

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/json/MultilabelModelReader.scala
```diff
@@ -0,0 +1,111 @@
...
+    refInfoK: RefInfo[K],
+    jsonFormatK: JsonFormat[K],
+    serEvK: SerializabilityEvidence[K])
+   extends JsonReader[MultilabelModel[U, K, A, B]]
+      with MultilabelModelJson
+      with EitherHelpers
+      with RegFeatureCompiler { self =>
+
+  /**
+    * Creates a reader for [[MultilabelModel]] that has a less specific type assuming that we
```
jmorra, line 46: I feel like I've read this comment before...

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/package.scala
```diff
@@ -0,0 +1,75 @@
...
+    * Sparse features related to the labels.  Other outer sequence aligns with the `Labels` and `LabelIndices`
+    * sequences meaning `SparseLabelDepFeatures[i]` relates to the features of `Labels[i]`.
+    */
+  private[multilabel] type SparseLabelDepFeatures = Labels[SparseFeatures]
+
+  /**
+    * A sparse multi-label predictor takes:
+    *
+    - features
+    - labels for which a prediction should be produced
```
jmorra, line 49: What happens if the labels for which a prediction should be produced contains a label that was not known at train time?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/package.scala
```diff
@@ -0,0 +1,75 @@
...
+    (SparseFeatures, Labels[K], LabelIndices, SparseLabelDepFeatures) => Try[Map[K, Double]]
+
+  /**
+    * A lazy version of a sparse multi-label predictor.  It is a curried zero-arg function that
+    * produces a sparse multi-label predictor.
+    *
+    * This definition is "lazy" because we can't guarantee that the underlying predictor is
+    * `Serializable` so we pass around a function that can be cached in a ''transient''
+    * `lazy val`.  This function should however be `Serializable` and testing should be done
+    * to ensure that each predictor producer is `Serializable`.
```
jmorra, line 67: Is this testing done?

aloha-core/src/main/scala/com/eharmony/aloha/reflect/RuntimeClasspathScanning.scala
```diff
@@ -0,0 +1,66 @@
...
+    *           `methodName`
+    * @return a sequence of `A` instances that could be found.
+    */
+  protected[this] def scanObjects[OBJ: ClassTag, A: ClassTag](
+      methodName: String,
+      packageToSearch: String = aloha.pkgName
+  ): Seq[A] = {
+    val reflections = new Reflections(aloha.pkgName)
+    import scala.collection.JavaConversions.asScalaSet
+    //    val classA = implicitly[ClassTag[A]].runtimeClass
```
jmorra: Dead code.

aloha-core/src/main/scala/com/eharmony/aloha/reflect/RuntimeClasspathScanning.scala
```diff
@@ -0,0 +1,66 @@
...
+import com.eharmony.aloha
+import org.reflections.Reflections
+
+import scala.reflect.ClassTag
+import scala.util.Try
+
+/**
+  * Created by ryan.deak on 9/6/17.
+  */
+trait RuntimeClasspathScanning {
```
jmorra, line 12: Can this not be done using `org.reflections`?

aloha-core/src/test/scala/com/eharmony/aloha/dataset/RowCreatorProducerTest.scala
```diff
@@ -41,3 +47,9 @@ class RowCreatorProducerTest {
         }
     }
 }
+
+object RowCreatorProducerTest {
+    private val WhitelistedRowCreatorProducers = Set[Class[_]](
+        classOf[VwMultilabelRowCreator.Producer[_, _]]
+    )
+}
```
jmorra: newline

aloha-core/src/test/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreatorTest.scala
```diff
@@ -0,0 +1,228 @@
...
+  private val LabelsInTrainingSet = Vector("zero", "one", "two")
+  private val NegDummyClass = Int.MaxValue.toLong + 1
+  private val PosDummyClass = NegDummyClass + 1
+  private val PosVal = -1
+  private val NegVal = 0
+  private val X = Map.empty[String, Any]
+  private val SharedPrefix = "shared "
+
+  private val DummyLabels = List(
+    s"$NegDummyClass:$NegVal |y N",
```
jmorra: I would encourage you to not write `N` and `P` here, but to use the constants you created in the main code.

aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwMultilabelParamAugmentation.scala
```diff
@@ -0,0 +1,453 @@
...
+
+  protected type VWNsSet = Set[Char]
+  protected type VWNsCrossProdSet = Set[(Char, Char)]
+
+  /**
+    * Adds VW parameters to make the parameters work as an Aloha multilabel model.
+    *
+    * The algorithm works as follows:
+    *
+    1. Ensure the VW `csoaa_ldf` or `wap_ldf` reduction is specified in the supplied VW
```
jmorra, line 28: Should these all be 1?

aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/VwJniModel.scala
```diff
@@ -239,21 +240,13 @@ object VwJniModel
...
+              val (nss, defaultNs, missing) =
+                allNamespaceIndices(names, vw.namespaces.getOrElse(sci.ListMap.empty))
 
-              val vwParams = vw.vw.params.fold("")(_.fold(_.mkString(" "), x => x)).trim
-
-              val defaultNs = (indices.keySet -- (Set.empty[String] /: nssRaw)(_ ++ _._2)).flatMap(indices.get).toList
+              if (missing.nonEmpty)
+                info(s"Ignoring features in namespaces not in feature list: $missing")
 
+              val vwParams = vw.vw.params.fold("")(_.fold(_.mkString(" "), x => x)).trim
```
jmorra, line 249: Do you need to add a `json` function to this class to handle multilabel models?

aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwMultilabelParamAugmentation.scala
```diff
@@ -0,0 +1,453 @@
...
+    * )
+    * }}}
+    *
+    * @param vwParams current VW parameters passed to the VW JNI
+    * @param namespaceNames it is assumed that `namespaceNames` is a superset
+    *                       of all of the namespaces referred to by any flags
+    *                       found in `vwParams`.
+    * @return
+    */
+  def updatedVwParams(vwParams: String, namespaceNames: Set[String]): Either[VwParamError, String] = {
```
jmorra: I love the amount of docs here, very clear.

jmorra: Can you add support for `--ring_size` too.

aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwMultilabelParamAugmentation.scala
```diff
@@ -0,0 +1,453 @@
...
+/**
+  * Created by ryan.deak on 10/5/17.
+  */
+protected trait VwMultilabelParamAugmentation {
+
+  protected type VWNsSet = Set[Char]
+  protected type VWNsCrossProdSet = Set[(Char, Char)]
+
+  /**
+    * Adds VW parameters to make the parameters work as an Aloha multilabel model.
```
jmorra, line 24: I would add a comment here about how this is experimental and tied to a specific version of VW.

aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwMultilabelParamAugmentation.scala
```diff
@@ -0,0 +1,453 @@
...
+    val qs = il.flatMap(n =>
+      if (oldIgnored.contains(n) || oldIgnoredLinear.contains(n)) Nil
+      else List(s"${labelNs.labelNs}$n")
+    )
+
+    // Turn quadratic into cubic and cubic into higher-order interactions.
+    val cs  = createLabelInteractions(oldInteractions, oldIgnored, labelNs, _ == 2)
+    val hos = createLabelInteractions(oldInteractions, oldIgnored, labelNs, _ >= 3)
+
+    val quadratics = qs.toSeq.sorted.map(q => s"-q$q" ).mkString(" ")
```
jmorra, line 281: You can probably making this simpler by just making `-q`, `--cubic`, and `--interactions` all go to `--interactions` directly.

aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwSparseMultilabelPredictor.scala
```diff
@@ -0,0 +1,182 @@
...
+  * @param params VW parameters.
+  * @param defaultNs The list of indices into the `features` sequence that does not have
+  *                  an exist in any value of the `namespaces` map.
+  * @param namespaces Mapping from namespace name to indices in the `features` sequence passed
+  *                   to the `apply` method.  There should be no empty namespaces, meaning
+  *                   ''key-value'' pairs appearing in the map should not values that are empty
+  *                   sequences.  '''This is a requirement.'''
+  * @tparam K the label or class type.
+  */
+// TODO: Comment this function.  It requires a lot of assumptions.  Make those known.
```
jmorra: Let's do this now.

aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwSparseMultilabelPredictor.scala
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  /**
+    * Get the VW parameters used to invoke the underlying VW model.
+    * @return VW parameters.
+    */
+  def vwParams(): String = updatedParams
+
+  /**
+    * Given the input, form a VW example, and delegate to the underlying ''CSOAA LDF'' VW model.
```
jmorra, line 69: Not here, but in general we should say why we MUST use a CSOAA LDF model and not just a CSOAA model when that could be sufficient.  We should comment on how VW doesn't export those types correctly through the JNI and how that's the fault not of the JNI, but of the CSOAA class on the C side.  We should also comment on how we'll add more support that for CSOAA which is much simpler once that work is done within VW.

aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwSparseMultilabelPredictorProducer.scala
```diff
@@ -0,0 +1,53 @@
+package com.eharmony.aloha.models.vw.jni.multilabel
+
+import com.eharmony.aloha.io.sources.ModelSource
+import com.eharmony.aloha.models.multilabel._
+import com.eharmony.aloha.models.vw.jni.multilabel.json.VwMultilabelModelPluginJsonReader
+import com.eharmony.aloha.reflect.RefInfo
+import spray.json.{JsonFormat, JsonReader}
+
+/**
+  * A thing wrapper responsible for creating a [[VwSparseMultilabelPredictor]].  This
```
jmorra: `A thin wrapper`...

aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/multilabel/json/VwMultilabelModelPluginJsonReader.scala
```diff
@@ -0,0 +1,69 @@
...
+
+    if (missing.nonEmpty)
+      info(s"features in namespaces not found in featureNames: $missing")
+
+    val namespaceNames: Set[String] = namespaces.map(_._1)(breakOut)
+    val labelAndDummyLabelNss = determineLabelNamespaces(namespaceNames)
+
+    labelAndDummyLabelNss match {
+      case Some(LabelNamespaces(labelNs, _)) =>
+        // TODO: Should we remove this. If not, it must contain the --ring_size [training labels + 10].
```
jmorra: I think so.

aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/multilabel/json/VwMultilabelModelPluginJsonReader.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  private[multilabel] def vwParams(params: Option[Either[Seq[String], String]]): String =
+    params.fold("")(e => e.fold(ps => ps.mkString(" "), identity[String])).trim
+
+  private[multilabel] def notObjErr(json: JsValue): String = {
+    val str = json.prettyPrint
+    val substr = str.substring(0, JsonErrStrLength)
+    s"JSON object expected.  Found " + substr + (if (str.length != substr.length) " ..." else "")
+  }
+}
```
jmorra, line 64: newline

aloha-vw-jni/src/test/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwMultilabelModelTest.scala
```diff
@@ -0,0 +1,508 @@
...
+
+  private implicit def vecWriter[K: JsonWriter]: RootJsonFormat[Vector[K]] =
+    DefaultJsonProtocol.vectorFormat(DefaultJsonProtocol.lift(implicitly[JsonWriter[K]]))
+
+  private[multilabel] def modelJson[K: JsonWriter](
+      modelSource: ModelSource,
+      labelsInTrainingSet: Vector[K],
+      labelsOfInterest: Option[String] = None) = {
+
+//    implicit val vecWriter = vectorFormat(DefaultJsonProtocol.lift(implicitly[JsonWriter[K]]))
```
jmorra: Dead code.

aloha-vw-jni/src/test/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwMultilabelParamAugmentationTest.scala
```diff
@@ -0,0 +1,287 @@
...
+              assertEquals(Set('y'), ignoredNs)
+              labelNs.labelNs
+          }
+
+        assertEquals('Y', ignored)
+    }
+  }
+}
+
+object VwMultilabelParamAugmentationTest {
```
jmorra, line 286: Remove

aloha-vw-jni/src/test/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwSparseMultilabelPredictorTest.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  private def checkVwRingSize(numLabelsInTrainingSet: Int, ringSize: Int): Unit = {
+    assertEquals(
+      "vw --ring_size parameter is incorrect:",
+      numLabelsInTrainingSet + VwSparseMultilabelPredictor.AddlVwRingSize,
+      ringSize.toInt
+    )
+  }
+}
```
jmorra, line 73: newline

jmorra (changes requested):
Looks really nice, left some comments.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
ryan-deak-zefr: Emphasis.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
ryan-deak-zefr: Yeah.  I like that.  Let me think on that for a few and come up with something.

ryan-deak-zefr: @jmorra I think the requirement that "*negative labels added need to change with EVERY line*" is important, but I don't think it fits into the current architecture.  It seems like you are (at least implicitly) advocating two things:

1. Intentionally making `apply` "*more*" impure, and definitely *not* idempotent or a function (in the math sense).
1. Maintaining state inside `VwMultilabelRowCreator` (by virtue of *(1)*)

What if we have a training set that is distributed across multiple nodes.  We apply the `VwMultilabelRowCreator` to the elements on each node and combine the results?  If we want to do what I think you are suggesting, I think we'll need to make some pretty big design changes.  I suggest the following:

1. We'll need a constructor `seed` parameter.  This could be a value, but if so, the calling code would be responsible for varying the seed for each node on which the `VwMultilabelRowCreator` is run.  It could also be an impure and side-effecting function is idempotent (in context), meaning it would produce the same result every time it is run on the same node.  This would allow the seed *generator* parameter to be set once and called on each node to produce the seed for that node. 
   ```scala
   seed: () => Long // Not necessarily a long but something to seed pseudo-randomness.
    ```
1. We'll need a new interface like `StatefulRowCreator`.  This interface will use something like the `State` monad in [cats](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/package.scala#L41-L42) or [scalaz](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/package.scala#L147-L156).  I would recommend stack safety because its implementation stack-safe (remember the `tut` bug in 0.5.3?).
1. We'll have to use the `StatefulRowCreator` differently most likely, we could expose two methods, one that operates on single items and requires the caller the maintain the state and pass it in.  The other method could operate on multiple elements (collections, iterators) and hide the state information.

What do you think about this stuff?  I think if we try to jam this type of mutability into the current interface, it'll most likely produce disastrous results.

ryan-deak-zefr (commented):
@jmorra. What do you think about this?

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
ryan-deak-zefr: @jmorra: I am proposing the following trait:

```scala
trait StatefulRowCreator[-A, +B, S] extends Serializable {
  // An initial state derived from the constructor
  def initialState: S

  // Caller maintains state.
  def apply(a: A, s: S): ((MissingAndErroneousFeatureInfo, Option[B]), S)

  // Iterator doesn't return state because it's non-strict.  Output Iterator length
  // is same as input iterator length.  May be infinite length.  This should be a constant
  // time operation.
  def apply(as: Iterator[A], s: S): Iterator[(MissingAndErroneousFeatureInfo, Option[B])]

  // Output Vector length is same as input length.  Returns the final state.
  // This is a linear time operation.
  def apply(as: Vector[A], s: S): (Vector[(MissingAndErroneousFeatureInfo, Option[B])], S)
}
```

I think it was a mistake in retrospect to not have `Option[B]` in the `RowCreator` trait.  This should be rectified here and we can address the `Option` in the `RowCreator` elsewhere.

I'm happy to change the function names but I like the signatures.  (`map` maybe?). 

We could consider putting the state as the first parameter at least in the Iterator and Vector methods so that we can curry and supply the initial state to get back a function of one argument.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
ryan-deak-zefr: `.toSeq` is on an option.  It could be removed without effect due to the implicit function that changes options to iterables.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
ryan-deak-zefr: done

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
deaktator: Requires class instantiation.  Changing to `case s if s.length != 0 => `.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
deaktator: Sure

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
deaktator: I don't understand.  Are you saying that `positiveLabelIndices` is not guaranteed to be ***O***(**1**)?  That's definitely true, but neither is `Set.contains`.  See [Collections Performance Characteristics](http://docs.scala-lang.org/overviews/collections/performance-characteristics.html).  I don't think this is an issue as it's a package private function that shouldn't be called outside Aloha.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreator.scala
deaktator: This is more performant than an immutable implementation.  That comment was in reference to mutability, not performance.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: Do you believe we can jam TensorFlow into this?  Is it sufficiently general?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: [com.eharmony.aloha.ModelSerializabilityTest](https://github.com/eHarmony/aloha/blob/master/aloha-core/src/test/scala/com/eharmony/aloha/ModelSerializabilityTest.scala) should take care of this.  There are also a bunch of tests in `MultilabelModelTest`.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: Not worrying about it.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: Don't think it will be.  We can address this later if there's a size issue.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModel.scala
deaktator: Once per prediction.  How would you speed it up?

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/MultilabelModelParserPlugin.scala
deaktator: done

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/PluginInfo.scala
deaktator: *I* ❤️ *the performance of* `get` *and* `apply` *on* `ListMap`!  Also, it provides a way to index on key-value pairs and maintain iteration order based on insertion order.  It also allows us to read the features as a map in the JSON.  For instance:

```javascript
{
  "modelType": "multilabel-sparse",
  "modelId": { "id": 1, "name": "NONE" },
  "numMissingThreshold": 0,
  "features": {
    "feature": "1"                // <== JSON Map maintains order in code.
  },
  "labelsInTrainingSet": [4, 8],
  "underlying": {
    "type": "vw",
    "modelSource": { "model": "..." },
    "namespaces": { "X": [ "feature" ] }
  }
}
```

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/json/MultilabelModelJson.scala
deaktator: done

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/package.scala
deaktator: It will not be in the output but will be in `LabelsAndInfo.labelsNotInTrainingSet` returned by `MultilabelModel.labelsForPrediction`.  This `LabelsAndInfo` instance is accepted by the error functions and success functions and these are reported in the Aloha audit trail.

aloha-core/src/main/scala/com/eharmony/aloha/models/multilabel/package.scala
deaktator: `VwSparseMultilabelPredictorTest.testSerializability`

aloha-core/src/main/scala/com/eharmony/aloha/reflect/RuntimeClasspathScanning.scala
deaktator: That's what I'm doing in that class.  What do you mean?

aloha-core/src/test/scala/com/eharmony/aloha/dataset/vw/multilabel/VwMultilabelRowCreatorTest.scala
deaktator: I don't think that's a good idea.  This is testing the constants are the expected values.  This is important, especially for the `PosVal` and `NegVal`. I will extract it though, like I did with the other variables.

aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/VwJniModel.scala
deaktator: One exists for [com.eharmony.aloha.models.vw.jni.multilabel.VwMultlabelJsonCreator](https://github.com/eHarmony/aloha/blob/163-multilabel/aloha-vw-jni/src/main/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwMultlabelJsonCreator.scala).  Do you think that most of that code should be refactored to be placed in MultlabelJsonCreator and make it accept an implicit way of parameterizing it based on PredictorProducer?

deaktator (commented):
@jmorra I think this addresses all of your PR comments.  I updated the downsampling to operate on 2<sup>31</sup> - 1 labels.

aloha-core/src/main/scala/com/eharmony/aloha/util/rand/Rand.scala
```diff
@@ -0,0 +1,120 @@
...
+            } while (bits - value + (i - 1) < 0)
+
+            value
+          }
+
+        // This is the key to maintaining the proper probabilities in the reservoir sampling.
+        // Wikipedia does a good job describing this in the proof by induction in the
+        // explanation for Algorithm R.
+        if (reservoirSwapInd < k)
+          reservoir(reservoirSwapInd) = (i - 1).toShort
```
deaktator: Need to remove "toShort"

aloha-core/src/main/scala/com/eharmony/aloha/dataset/StatefulRowCreator.scala
```diff
@@ -0,0 +1,123 @@
...
+    * }}}
+    *
+    * @param as Note the first element of `as` ''will be forced'' in this method in order
+    *           to construct the output.
+    * @param state the initial state to use at the start of the iterator.
+    * @return an iterator containing the `a` mapped to a
+    *         `(MissingAndErroneousFeatureInfo, Option[B])` along with the resulting
+    *         state that is created in the process.
+    */
+  def mapIteratorWithState(as: Iterator[A], state: S): Iterator[((MissingAndErroneousFeatureInfo, Option[B]), S)] = {
```
deaktator: @jmorra, do you think I should extract this to an object:

```scala
package com.eharmony.aloha.util

object MappingUtils {
  def mapIteratorWithState[A, B, S](as: Iterator[A], state: S)(f: (A, S) => (B, S)): Iterator[(B, S)] = {
    // implementation filled in
  }

  def mapSeqWithState[A, B, S](as: Seq[A], state: S)(f: (A, S) => (B, S)): (Vector[B], S) = {
    // implementation filled in
  }
}
```

And then have a trait:

```scala
trait MapWithState[A, B, S] { self: ((A, S) => (B, S)) => 
  def mapIteratorWithState(as: Iterator[A], state: S): Iterator[(B, S)] = 
    MappingUtils.mapIteratorWithState(as, state)(apply)

  def mapSeqWithState(as: Seq[A], state: S): (Vector[B], S) = 
    MappingUtils.mapSeqWithState(as, state)(apply)
}
```

deaktator (commented):
What do you think about refactoring `StatefulRowCreator` to make it useful elsewhere?

aloha-cli/src/test/scala/com/eharmony/aloha/cli/ModelTypesTest.scala
```diff
@@ -28,7 +28,8 @@ class ModelTypesTest {
       "ModelDecisionTree",
       "Regression",
       "Segmentation",
-      "VwJNI"
+      "VwJNI",
+      "multilabel-sparse"
```
jmorra: Why is the casing different here?

aloha-core/src/main/scala/com/eharmony/aloha/dataset/StatefulRowCreator.scala
jmorra: I don't understand why you don't model both of these functions together with a common interface such as `TraversableOnce`.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwDownsampledMultilabelRowCreator.scala
```diff
@@ -0,0 +1,378 @@
...
+  * @param seedCreator a "''function''" that creates a seed that will be used for randomness.
+  *                    The implementation of this function is important.  It should create a
+  *                    unique value for each unit of parallelism.  If for example, row
+  *                    creation is parallelized across multiple threads on one machine, the
+  *                    unit of parallelism is threads and `seedCreator` should produce unique
+  *                    values for each thread.  If row creation is parallelized across multiple
+  *                    machines, the `seedCreator` should produce a unique value for each
+  *                    machine.  If row creation is parallelized across machines and threads on
+  *                    each machine, the `seedCreator` should create unique values for each
+  *                    thread on each machine.  Otherwise, randomness will be striped which
```
jmorra, line 66: Finish your thought.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwDownsampledMultilabelRowCreator.scala
```diff
@@ -0,0 +1,378 @@
...
+    // The positiveLabelsFunction is invoked once and all labels are produced.
+    // Then a set is produced that is used to partition all of the labels in
+    // the training data into positive and negative.  It seems like a slightly
+    // more efficient way to do this would be to create a sorted array of positives
+    // indices.  Then a negative index array could be constructed with the appropriate
+    // size based on the size of allLabelsInTrainingSet and the size of the positives
+    // array.  Next allLabelsInTrainingSet.indices and the positiveIndices array could
+    // be iterated over simultaneously and if the current index in
+    // allLabelsInTrainingSet.indices isn't in the positiveIndices array, then it must
+    // be in the negativeIndices array.  Then then offsets into the two arrays are
```
jmorra: Then then

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwDownsampledMultilabelRowCreator.scala
```diff
@@ -0,0 +1,378 @@
...
+
+    ((missingAndErrs, Option(vwInput)), newSeed)
+  }
+}
+
+
+object VwDownsampledMultilabelRowCreator extends Rand {
+
+  // Expose initSeedScramble to companion class.
+  private def scramble(initSeed: Long): Long = initSeedScramble(initSeed)
```
jmorra, line 195: Why bother to have a function that is just a pass through?

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwDownsampledMultilabelRowCreator.scala
```diff
@@ -0,0 +1,378 @@
...
+    *         missing and error information and second element is an optional result.
+    *         The second element of the outer Tuple2 is the new state.
+    */
+  override def apply(a: A, seed: Long): ((MissingAndErroneousFeatureInfo, Option[Array[String]]), Long) = {
+    val (missingAndErrs, features) = featuresFunction(a)
+
+    // Get the lazy val once.
+    val labToInd = labelToInd
+
+    // TODO: This seems like it could be optimized.
```
jmorra, line 152: Should this comment be moved to the function called?

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwDownsampledMultilabelRowCreator.scala
```diff
@@ -0,0 +1,378 @@
...
+    // TODO: Write calling code to abstract over costs so knowledge of the costs isn't necessary.
+    val negWt =
+      if (numDownsampledLabels == negLabels) {
+        // No downsampling occurs.
+        NegativeCost.toString
+      }
+      else {
+        // Determine the weight for the downsampled negatives.
+        // If the cost of negative examples is positive, then the weight will be
+        // strictly greater than .
```
jmorra: greater than ???

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwDownsampledMultilabelRowCreator.scala
```diff
@@ -0,0 +1,378 @@
...
+    val (downsampledNegLabelInd, newSeed) =
+      sampleCombination(negLabels, numDownsampledLabels, seed)
+
+    val p = posLabelInd.size
+    val n = downsampledNegLabelInd.length
+
+    // This code is written with knowledge of the constant's value.
+    // TODO: Write calling code to abstract over costs so knowledge of the costs isn't necessary.
+    val negWt =
+      if (numDownsampledLabels == negLabels) {
```
jmorra, line 253: Should this be `==` or `<=`?

aloha-core/src/main/scala/com/eharmony/aloha/util/rand/Rand.scala
```diff
@@ -0,0 +1,120 @@
...
+    * @param seed the seed to use for random selection
+    * @return a tuple 2 containing the array of 0-based indices representing
+    *         the ''k''-combination and a new random seed.
+    */
+  protected def sampleCombination(n: Int, k: Int, seed: Seed): (Array[Index], Seed) = {
+
+    // NOTE: This isn't idiomatic Scala code but it will operate in hot loops in minimizing
+    //       object creation is important.
+
+    if (n <= k) {
```
jmorra, line 69: Honestly, this is very difficult to read, but I'm going to assume that you did your research here judging by the comments.

aloha-vw-jni/src/test/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwMultilabelParamAugmentationTest.scala
```diff
@@ -283,5 +284,7 @@ class VwMultilabelParamAugmentationTest extends VwMultilabelParamAugmentation {
 }
 
 object VwMultilabelParamAugmentationTest {
-
+  val DefaultNumLabels = 0
+  val DefaultRingSize =
+    s"--ring_size ${DefaultNumLabels + VwSparseMultilabelPredictor.AddlVwRingSize}"
 }
```
jmorra: newline

aloha-vw-jni/src/test/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwMultilabelDownsampledModelTest.scala
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+  * Created by ryan.deak on 11/6/17.
+  */
+@RunWith(classOf[BlockJUnit4ClassRunner])
+class VwMultilabelDownsampledModelTest {
+
+  import VwMultilabelDownsampledModelTest.EndToEndDatasetSpec
+
+  @Test def test(): Unit = {
```
jmorra, line 33: I don't see a test for the actual VW input.  Is that here?

jmorra (approved):
I'm going to approve this with some comments that may not require any code changes.  Let me know

aloha-core/src/main/scala/com/eharmony/aloha/dataset/StatefulRowCreator.scala
deaktator: @jmorra what do you think about something like this:

```scala
import scala.collection.SeqLike
import scala.collection.generic.CanBuildFrom
import scala.collection.breakOut
import scala.collection.{immutable => sci}

def statefulMap[A, B, S, Repr <: sci.Seq[A], That]
    (as: SeqLike[A, Repr], s: S)
    (f: (A, S) => (B, S))
    (implicit cbf: CanBuildFrom[Repr, (B, S), That]): That = {
  if (as.isEmpty)
    cbf().result()
  else {
    val initEl = f(as.head, s)
    as.tail.scanLeft(initEl){ case ((_, newS), a) => f(a, newS) }(breakOut)
  }
}
```

Then we can use any immutable sequence like you suggested and we can maintain the sequence's structure by not supplying a `scala.collection.generic.CanBuildFrom` or we can change the structure by supplying a `CanBuildFrom`.  See below:

```scala
val sCycleModulo4 = Stream.iterate(0)(s => (s + 1) % 4)
val vCycleModulo4 = Vector.iterate(0, 10)(s => (s + 1) % 4)
val addAndIncrement = (a: Int, s: Int) => ((a + s).toDouble, s + 1)

val stream = statefulMap(sCycleModulo4, 0)(addAndIncrement)

val list: List[(Double, Int)] = 
  statefulMap(vCycleModulo4, 0)(addAndIncrement)(breakOut)
```

We also get safety guarantees for things that are not immutable (e.g. `Array`).  For instance:

```scala
val aCycleModulo4 = Array.iterate(0, 10)(s => (s + 1) % 4)
statefulMap(aCycleModulo4, 0)(addAndIncrement)
```

Returns:

```
<console>:21: error: inferred type arguments [Int,Nothing,Int,Array[Int],Nothing] 
do not conform to method statefulMap's type parameter bounds 
[A,B,S,Repr <: scala.collection.immutable.Seq[A],That]

statefulMap(aCycleModulo4, 0)(addAndIncrement)
^
```

I chose `sc.SeqLike` and `sci.Seq` over `TraversableLike` and `Traversable` or `GenTraversableLike` and `GenTraversable` for a few reasons.  The first is that there are no iteration order guarantees for `Traversable`s so the function wouldn't be guaranteed to be idempotent (which I didn't like), whereas sequences do.  The reason I chose `Repr <: sci.Seq[A]` is that I like the idea that of immutability.  A third reason to choose immutable is that it precludes arrays.  If we want arrays, there should be a separate method because we would have to do an ***O***(***N***) array copy to get the tail and another ***O***(***N***) copy to produce the result.  This is both slow in terms of runtime and excessive in terms of memory usage.

aloha-cli/src/test/scala/com/eharmony/aloha/cli/ModelTypesTest.scala
deaktator: Yeah.  I want opinions on the name here.  I am neither set on, nor particularly enamored with the name name `"multilabel-sparse"`.  Suggestions?

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwDownsampledMultilabelRowCreator.scala
deaktator: I don't think so because the computation of `positiveIndices` is "*part of the algorithm*".

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwDownsampledMultilabelRowCreator.scala
deaktator: It needs it because of the way that Scala does mix-in composition.  If you removed it and tried to call `initSeedScramble` directly, the calling code wouldn't be able to access it.  `Rand` could have been mixed into both the class and companion object but I didn't like that either.

aloha-core/src/main/scala/com/eharmony/aloha/dataset/vw/multilabel/VwDownsampledMultilabelRowCreator.scala
deaktator: `==`.  Since `numDownsampledLabels == math.min(numNegLabelsTarget, negLabels)`, it is certain that `numDownsampledLabels <= negLabels`.  We only don't downsample when `numDownsampledLabels == negLabels`.

aloha-core/src/main/scala/com/eharmony/aloha/util/rand/Rand.scala
deaktator: You did say you wanted "*fast*" and that hashing would be too slow.  This is what happens when making deals with the devil.  And the devil is *always* in the details.  *Muah ha ha ha!*

aloha-vw-jni/src/test/scala/com/eharmony/aloha/models/vw/jni/multilabel/VwMultilabelDownsampledModelTest.scala
deaktator: Yeah.  Sort of.  The VW input is generated implicitly and the prediction output is produced and tested against the marginal probabilities.  So, it works implicitly.

jmorra (approved):

PR #196, opened 2017-11-08
Preparing for 5.1.0

deaktator:


jmorra (approved):
I'm assuming these are the changes I've already approved. If this is not the case please let me know.

PR #204, opened 2017-11-28
Closes #203

### LINKED ISSUE #203: Add unit tests to confirm feature functions don't call referenced functions when data is missing

deaktator:
## Summary

There was some discussion as to whether feature function bodies are only called when data is present.  While this has always been the design, I don't think there are ever a test for this.  This was added to ***ensure*** that's the case. 

## Resolves
* Provides guarantees related to #203

## Tests
See `CompiledSemanticsTest.testFeatureFunctionsDoNotCallReferencedFunctionsWhenDataIsMissing` in *aloha-core*.  Also adds support for doing a similar thing with other `CompiledSemanticsPlugin`s by creating a trait that can be mixed into test class companion objects.

## Code Reviewer(s)
@jmorra

### DIFF
--- a/aloha-core/src/test/scala/com/eharmony/aloha/semantics/LazyInitSupport.scala
+++ b/aloha-core/src/test/scala/com/eharmony/aloha/semantics/LazyInitSupport.scala
@@ -1,0 +1,52 @@
+package com.eharmony.aloha.semantics
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+  * Provides support for testing that feature functions produced by semantics
+  * do not call the feature function body when optional data is not present.
+  *
+  * This should probably be mixed into unit test companion objects.
+  *
+  * @author deaktator
+  * @since 11/28/2017
+  */
+trait LazyInitSupport {
+  private[this] val NotCalled = false
+  private[this] val Called = true
+
+  /**
+    * Only to be modified by `wasCalled`.
+    */
+  private[this] val indicator = new AtomicBoolean(NotCalled)
+
+  /**
+    * Attempt to set `indicator` to `true` assuming `indicator` wasn't already set.
+    * `true` if indicator wasn't previously set; otherwise, `false`.
+    */
+  private[this] lazy val wasCalled = indicator.compareAndSet(NotCalled, Called)
+
+  /**
+    * This function should be called in a feature function created by the
+    * semantics to test that when `a` represents optional data that is not
+    * present, the `lazy val` is not initialized.  For instance:
+    *
+    * {{{
+    * val semantics: Semantics[Option[String]] = ???
+    * val spec = "code(${missingVariable})"
+    * val default = Some(None)
+    * val f: GenAggFunc[A, Option[Boolean]] =
+    *   semantics.createFunction[Option[Boolean]](spec, default).right.get
+    * val y = f(None)
+    * assertFalse(codeWasCalled)
+    * }}}
+    * @param a some value.
+    * @return `true`, ''assuming'' `indicator` ''wasn't modified''.
+    */
+  def code(a: Any): Boolean = wasCalled
+
+  /**
+    * @return whether `code` was called.
+    */
+  def codeWasCalled: Boolean = indicator.get
+}
--- a/aloha-core/src/test/scala/com/eharmony/aloha/semantics/LazyInitSupport.scala
+++ b/aloha-core/src/test/scala/com/eharmony/aloha/semantics/LazyInitSupport.scala
@@ -11,20 +11,20 @@ import java.util.concurrent.atomic.AtomicBoolean
   * @author deaktator
   * @since 11/28/2017
   */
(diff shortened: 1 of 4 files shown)


aloha-core/src/test/scala/com/eharmony/aloha/semantics/LazyInitSupport.scala
```diff
@@ -0,0 +1,52 @@
...
+  * @since 11/28/2017
+  */
+private[aloha] abstract class LazyInitSupport {
+  private val NotCalled = false
+  private val Called = true
+
+  /**
+    * Only to be modified by `wasCalled`.
+    */
+  private val indicator = new AtomicBoolean(NotCalled)
```
deaktator, line 21: privacy is important here, since we don't want the sub type to modify these.

aloha-core/src/test/scala/com/eharmony/aloha/semantics/LazyInitSupport.scala
```diff
@@ -0,0 +1,52 @@
...
+/**
+  * Provides support for testing that feature functions produced by semantics
+  * do not call the feature function body when optional data is not present.
+  *
+  * This should probably be mixed into unit test companion objects.
+  *
+  * @author deaktator
+  * @since 11/28/2017
+  */
+private[aloha] abstract class LazyInitSupport {
```
deaktator, line 14: Making this an abstract class locks down the privacy of the members so they can't be accessed in sub-classes without trickery.

aloha-core/src/test/scala/com/eharmony/aloha/semantics/LazyInitSupport.scala
```diff
@@ -0,0 +1,52 @@
...
+  /**
+    * Only to be modified by `wasCalled`.
+    */
+  private val indicator = new AtomicBoolean(NotCalled)
+
+  /**
+    * Attempt to set `indicator` to `true` assuming `indicator` wasn't already set.
+    * `true` if indicator wasn't previously set; otherwise, `false`.
+    */
+  private lazy val wasCalled = indicator.compareAndSet(NotCalled, Called)
```
deaktator, line 27: This was added to check that lazy vals work. If lazy vals, didn't work, that would likely be a Scala language problem, but concerns were raised about this in conversations, so I felt it prudent to add.

aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/plugin/AnyNameMissingValueSemanticsPlugin.scala
```diff
@@ -0,0 +1,26 @@
...
+  * A semantics whose variables (''with any name'') always return missing data.
+  * @param refInfoA reflection information about `A`
+  * @param evOpt evidence that `A` is an `Option` of some kind.
+  * @tparam A the input type to a model (an `Option` of some kind).
+  * @author deaktator
+  * @since 11/28/2017
+  */
+private[aloha] class AnyNameMissingValueSemanticsPlugin[A](implicit
+    val refInfoA: RefInfo[A],
+    evOpt: A <:< Option[_])
```
deaktator, line 16: Ensures `A` is an `Option`.

aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/plugin/AnyNameMissingValueSemanticsPlugin.scala
```diff
@@ -0,0 +1,26 @@
...
+  extends CompiledSemanticsPlugin[A] {
+
+  /**
+    * Produces code for a function that takes an `Option[A]` and returns an `Option.empty[A]`.
+    * @param spec an '''ignored''' string-based specification of a function.
+    * @return
+    */
+  override def accessorFunctionCode(spec: String): Right[Nothing, OptionalAccessorCode] =
+    Right(OptionalAccessorCode(Seq(s"(o: ${RefInfoOps.toString[A]}) => o.filterNot(_ => true)")))
```
deaktator, line 25: Same as 

```scala
(o: Option[A]) => Option.empty[A]
```

aloha-core/src/test/scala/com/eharmony/aloha/semantics/compiled/CompiledSemanticsTest.scala
```diff
@@ -108,6 +143,8 @@ class CompiledSemanticsTest {
     }
 }
 
+object CompiledSemanticsTest extends LazyInitSupport
```
deaktator, line 146: Added the `code` function to the companion object so that it can be used in the spec in the test to determine if `code` is called.

jmorra (changes requested):
We know that this isn't true, how does this reconcile that problem?

deaktator:
Closing.  This provides no real value.

